PR #6, opened 2022-10-12
WS: Comment out Baroque WS patch

JordanTheToaster:
Appears to break both software and hardware rendering.

Screenshots from RedDevilus

image

image

### DIFF
--- a/cheats_ws/0CFFFBCC.pnach
+++ b/cheats_ws/0CFFFBCC.pnach
@@ -1,14 +1,16 @@
 gametitle=Baroque (PAL-E) (SLES-55216)
 comment=Widescreen hack by ElHecht
 
+//Apepars to totally break rendering in both hardware and software currently.
+
 // 16:9
-patch=1,EE,0039f504,word,3c033f40 // 3c033f80 hor fov
-patch=1,EE,0039f50c,word,4483f000 // 44832800
-patch=1,EE,0039f510,word,461ef143 // 00000000
-patch=1,EE,0039f514,word,4602f083 // 46022883
-patch=1,EE,0044cf30,word,3f5eb852 // 3f947ae1 subtitle hor fov
-patch=1,EE,00166ff0,word,3c054028 // 3c054060 hud hp/vt
-patch=1,EE,00167030,word,3c053ff0 // 3c054020 hud hp/vt
+//patch=1,EE,0039f504,word,3c033f40 // 3c033f80 hor fov
+//patch=1,EE,0039f50c,word,4483f000 // 44832800
+//patch=1,EE,0039f510,word,461ef143 // 00000000
+//patch=1,EE,0039f514,word,4602f083 // 46022883
+//patch=1,EE,0044cf30,word,3f5eb852 // 3f947ae1 subtitle hor fov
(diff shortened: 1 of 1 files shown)


Aced14:
I believe the same issue was recently identified and resolved in the NTSC-U version of that widescreen pnach. Maybe the NTSC-U fix could be ported to the PAL/NTSC-J versions?

**Excerpts from the PCSX2 widescreen game patches forum thread...**

> **[Rosenkreutz (reply #7.485):](https://forums.pcsx2.net/Thread-PCSX2-Widescreen-Game-Patches?pid=627096#pid627096)**
>  I noticed an issue with the widescreen patch for the game, Baroque. The game normally features a grain filter in addition to a ghosting effect around the characters which is used both for atmosphere and to obscure the fact that some NPC models only fully load in when you approach them. Image <span>#</span>1 is what the game is supposed to look like and Image <span>#</span>2 is how it looks with the widescreen patch on, as you can see the widescreen patch disables both of the grain filter and the ghosting effect. Someone on the PCSX2 Discord was able to edit the patch so that the grain filter was still present while using it (Image <span>#</span>3,) but they were unable to do the same for the ghosting effect. Would it be possible to edit the patch so that the ghosting effect would still be present? Based on what I was told it sounds like the only reason that it isn't there is because the patch currently doesn't adjust the dimensions of the effect to suit the wider field of view. If that is indeed the case then I would imagine that it should be possible to get both the filter and ghosting effect working alongside the widescreen patch if said effect's dimensions were properly adjusted. Would anyone know how to go about doing this? 
> Normal (1) | Widescreen (2) | Widescreen + Grain Filter (3)
> --- | --- | ---
> Normal (1) | Widescreen (2) | Widescreen + Grain Filter (3)

---

> **[felixthecat1970 (reply #7.509):](https://forums.pcsx2.net/Thread-PCSX2-Widescreen-Game-Patches?pid=627592#pid627592)**
> **<span>@</span>Nemesis2000**
> hello, thanks many codes you post across this years Laugh,  so i ask if you can help this widescreen code fix for Baroque (USA) version, (keep me thinking some weeks after user post this problem) looks code posted in forums not apply to effects (dof- blur), i know effect can be disabled but i try to preserve effect for game atmosphere is more original that disable it; i look function who change the value to widescreen for (dof-blur) but after change value the effect is expanded vertical too, after 3F80 are some "float point operations" but i have not manage to correct or match  (sometimes effect is reduced, enlaged, mirrored etc) i leave some images and code explaination.
> Code:
> ```
> //Baroque (USA) SLUS_217.14;1) Game CRC = 0x4566213C
> patch=1,EE,00237914,extended,3C013F40 //3f80 function area for wide (Dof-Blur)
> //horizontal effect exetended ok but vertical too (images available with deafult widescreen codes)
> ```
> Images is from Baroque (USA version)
> 
> 
> widescreen-normal | widescreen-changed val
> --- | ---
> widescreen-normal | widescreen-changed val

---

> **[nemesis2000 (reply #7.511):](https://forums.pcsx2.net/Thread-PCSX2-Widescreen-Game-Patches?pid=627598#pid627598)**
> dof fix
> 
> patch=1,EE,0010F68C,word,4600BDC2

PR #37, opened 2023-02-28
Update 6FB69282.pnach

FuxorLuck:


refractionpcsx2:
Couldn't you have put all these in one PR instead of spamming the repo?

PR #47, opened 2023-03-02
Add more NTSC-U Ratchet 16:9 patches

icup321:
Adds 16:9 patch for the Greatest Hits version of Going Commando, and re-adds Up Your Arsenal 16:9 patch with an updated fixed version. Previously the UYA 16:9 code caused the boss in the first vid-comic level to break, but the author uploaded a fixed version, so there shouldn't be issues with it anymore afaik.


icup321:
I can now confirm that Up Your Arsenal is fully playable from start to finish with the 16:9 patch. No game-breaking issues happened to me, so I think it's perfectly safe to use this patch now.

refractionpcsx2:
Thanks for testing

PR #49, opened 2023-03-18
Pull in Latest WS Changes for Persona 3 FES USA

alexankitty:
Applies to: 94A82AAA.pnach
I updated the FMV fix to properly scale to ratio instead of cropping like Nemesis2000 did in the past. The community has also made a ton more changes since that need to be pulled in as well.
This also removes the persona artwork fixes as it's not a good fix and causes portraits to bend.


refractionpcsx2:
Can you not mix 480p and widescreen patches, please.

alexankitty:
Wasn't aware of that restriction, I will remove it.

alexankitty:
Corrected.

refractionpcsx2:
Thanks :) Yeah we don't really want to pollute widescreen patches with other stuff, same for no interlacing patches.

PR #51, opened 2023-03-18
WS: Fix for Onimusha Dawn of Dreams Pal version pnach

bigol83:
This is a fix for Onimusha Dawn of Dreams 812C5A96.pnach

The old one was broken and it only showed 2d elements, no 3d graphics.


refractionpcsx2:
Thanks :)

PR #69, opened 2023-05-18
Add 60FPS and DNAS patches

TheRhysWyrill:
- Adds every 60FPS patch from here: https://forums.pcsx2.net/Thread-60-fps-codes, formatted to be usable with PCSX2.

- Adds 55 DNAS patches to skip DNAS protection on games, allowing users to play online.

Should this PR be merged then a PR will also be made to PCSX2 to incorporate the use of these files.


refractionpcsx2:
I'd prefer the functionality to support this was done before this got merged, so we didn't have pointless patches/folders hanging around.

the 60FPS option can go with Widescreen/No interlacing
DNAS can go under the networking tab (if you can think of somewhere decent to put it), if not, with the Enable Cheats option.

kamfretoz (changes requested):
Haven't gotten to look through every single files, but some of them requires few fixes.

kamfretoz:
> I'd prefer the functionality to support this was done before this got merged, so we didn't have pointless patches/folders hanging around.
> 
> the 60FPS option can go with Widescreen/No interlacing DNAS can go under the networking tab (if you can think of somewhere decent to put it), if not, with the Enable Cheats option.

How about moving all the patches toggles  to the Emulation section and have all four of them (widescreen, no-interlace, 60 fps, dnas) all in one place?

Make a new section called "Patches" or something.

image

TheRhysWyrill:
If a new "Patches" section in Emulation would be preferable then I can do that, would be nice to have all the Patch toggles in the same place just for ease of use, I'll look through and fix the DNAS files when I get chance.

refractionpcsx2:
As the main PR has been closed due to the potential legal problems with DNAS patches, I'm gonna close this one for now.  It would be good in future if/when we redo the patches system that the 60fps patches get curated, since most of them provide a broken experience,  it's not good for the user, and it's not good for the poor guys who do the support.

But hopefully we can circle back to adding the 60fps patches in future though.

PR #88, opened 2023-06-14
5C891FF1.pnach

jucasleiva:
this is a patch for the game Black to run at 60 Fps

### DIFF
--- a/patches/5C891FF1.pnach
+++ b/patches/5C891FF1.pnach
@@ -1,0 +1,4 @@
+gametitle=BLACK (SLUS_21376)
+
+patch=1,EE,1040DF74,extended,00000001 // 60 fps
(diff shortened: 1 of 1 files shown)


stenzek:
This is not the correct format, it will not apply in PCSX2.

1. The files should now be named `SERIAL_CRC.pnach`.
2. The 60 fps patch should be under a named group.

Example:
```
gametitle=Title of Game

[60 FPS]
description=Increases the game's internal frame rate from 30 to 60 frames per second.
author=authors_name
patch=1,EE,....
```

Also, please provide confirmation that the game runs correctly. Many of these patches simply change the vsync rate, and the game runs at 2x speed, because they're internally fixed frame rate.

PR #90, opened 2023-06-14
8FCCB5D9.pnach

jucasleiva:
this is a patch for Ben 10 - Protector of Earth game to run at 60 Fps


stenzek:
This is not the correct format, it will not apply in PCSX2.

1. The files should now be named `SERIAL_CRC.pnach`.
2. The 60 fps patch should be under a named group.

Example:
```
gametitle=Title of Game

[60 FPS]
description=Increases the game's internal frame rate from 30 to 60 frames per second.
author=authors_name
patch=1,EE,....
```

PR #104, opened 2023-06-24
Patches for Asterix & Obelix XXL + Asterix & Obelix Kick Buttix

Martyste:
Updates the syntax, adds No-Interlacing codes, and replicates the Black Bars code from Kick Buttix to XXL.


refractionpcsx2:
Thanks :)

PR #116, opened 2023-07-22
Widescreen patch for Spyro A Hero's Tail PAL

CRASHARKI:
Ported the NTSC-U patch.


Mrlinkwii:
looks ok assuming it works

PR #120, opened 2023-07-27
Update and rename `664D4EAE.pnach` to `SLAJ-25060_664D4EAE.pnach`

israpps:
found matching ELF ID for this CRC on my own mastercodes archive

https://github.com/PS2-Widescreen/Bare-Mastercodes-bin


refractionpcsx2:
Thanks, I appreciate it!

PR #124, opened 2023-07-30
Patches: Add 50/60 FPS patches to Rayman Revolution, various Crash Ba…

CRASHARKI:
…ndicoot games and Spider-Man 2


Mrlinkwii:
were these tested ?

PR #127, opened 2023-07-31
Rename `SLKA_25051_0E6F7A7B.pnach` to `SLKA-25051_0E6F7A7B.pnach`

israpps:


refractionpcsx2:
Also good catch :D

israpps:
> Also good catch :D

say thanks to grep

this will be useful for you guys in the future

```cmd
cd patches
dir /b>LST.TXT
busybox.exe grep -Ev "[a-zA-Z]{4}-[0-9]{5}.*.pnach" LST.TXT
del LST.TXT
```

PR #138, opened 2023-08-16
Widescreen patch for WWF SmackDown! Just Bring It NTSC

J-McFc:
Patch was ported from ElHecht's PAL version about a year ago, haven't seen it added yet though.


refractionpcsx2:
Cool thanks :)

PR #141, opened 2023-08-19
Patches: Better widescreen fixes for 007: Agent Under Fire

CookiePLMonster:
This PR improves the current widescreen patch for 007: Agent Under Fire. The following improvements are now in:

1. Moved to a new naming scheme with an explicitly stated author.
2. Fixed the patch stomping on memory outside of the on-foot segments by making it apply conditionally. The patch was "lucky" that during the driving segments it stomped on some meaningless debug strings. Unfortunately, the game does not switch ELFs via BIOS, so it is not possible to split patches by CRC, and instead "if checks" need to be used.
3. Added a patch for the driving segments of the game that are handled by a separate ELF file.
4. Added support for the PAL version, all languages.

007 - Agent Under Fire_SLUS-20265_20230819002539
007 - Agent Under Fire_SLUS-20265_20230819002214

Drafted until I've discussed it with the dev of AUF Reloaded, since apparently there were past efforts in fixing the driving segments for widescreen - and I want to avoid missing any gotchas. Once it's un-drafted I'll also submit those patches on the forums.

### DIFF
--- a/patches/SLES-50539_992B46DC.pnach
+++ b/patches/SLES-50539_992B46DC.pnach
@@ -1,0 +1,30 @@
+gametitle=James Bond 007 - Agent Under Fire (E) (SLES-50539)
+
+[Widescreen 16:9]
+gsaspectratio=16:9
+description=16:9 widescreen patch
+author=Arapapa, Silent
+
+// Action segments
+// English
+patch=1,EE,D0320494,extended,01003F80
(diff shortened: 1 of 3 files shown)


CookiePLMonster:
Updated to fix misaligned manual aiming during the driving missions.

007 - Agent Under Fire_SLUS-20265_20230820124147

With this fixed, I think the patch is good to go.

PR #142, opened 2023-08-20
Create SLES-50385_A633A4BF

fox140cv:
Add hack widescreen region pal-s and patch 50 fps in scenes tested ok


Mrlinkwii:
can this be checked , it the skateboard stuff still works

PR #149, opened 2023-08-20
Update SLES-51054_ACB1989A.pnach

fox140cv:
the arapapa widescreen hack does not seem to do anything, I add this taken from the ELF that if it changes to 16:9. I also add the patch that runs at 50 FPS and tested up to the middle of the game


refractionpcsx2:
Duplicate?

fox140cv:
Maybe arapapa wrote that widescreen and it worked in the past, but you don't see any game change when activating that patch. Maybe I should replace the arapapa widescreen with mine, which does work

refractionpcsx2:
No no, I mean you PR'd it twice, it's conflicting, which means it just changed already.

fox140cv:
ok sorry i delete it

PR #157, opened 2023-08-20
Update SLES-50975_F6F9A91D.pnach

fox140cv:
Add patch unlock at 50/60fps tested and functional


patches/SLES-50975_F6F9A91D.pnach
```diff
@@ -9,3 +9,7 @@ patch=1,EE,0016EC34,word,3C023FE3 //3C023FAA (Increases hor. axis)
 patch=1,EE,0016EC38,word,34428E38 //3442AAAB
 
 
+[50/60 FPS]
```
refractionpcsx2, line 12: is it 50 or 60?

fox140cv: In both, both when selecting 50hz and when selecting 60hz

PR #183, opened 2023-08-20
Create SLES-52570_180F5C36.pnach

fox140cv:
add ported widescreen from other pal version and patch 50fps tested ok


fox140cv:
sorry it won't happen again

PR #184, opened 2023-08-20
Create SLES-50808_3AC738AA.pnach

fox140cv:
add parent crc for secondary elf to work when enabling patch

### DIFF
--- a/patches/SLES-50808_3AC738AA.pnach
+++ b/patches/SLES-50808_3AC738AA.pnach
@@ -1,0 +1,6 @@
+gametitle=Deus Ex [PAL-S] SLES-50808 3AC738AA. ELF secundary 7070DC2F
+
+[50 FPS]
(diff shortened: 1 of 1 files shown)


patches/SLES-50808_3AC738AA.pnach
```diff
@@ -0,0 +1,6 @@
+gametitle=Deus Ex [PAL-S] SLES-50808 3AC738AA. ELF secundary 7070DC2F
```
refractionpcsx2, line 1: It's spelt `secondary`

can you correct this please.

fox140cv: corrected, my apologies

refractionpcsx2: Thanks, I appreciate it.

PR #185, opened 2023-08-20
Create SCES-52033_59EB2A95.pnach

fox140cv:
add unlock patch to 50 FPS. tested ok


refractionpcsx2:
Can you not make a PR for every file please.

PR #188, opened 2023-08-23
Added 60 FPS patch to NFS Hot Pursuit 2 NTSC

J-McFc:


patches/SLUS-20362_1D2818AF.pnach
fox140cv, line 1: Could it be that you are missing the code?

J-McFc: Well yknow that would be it haha thank you.

patches/SLUS-20362_1D2818AF.pnach
```diff
@@ -7,3 +7,13 @@ comment=Widescreen Hack by nemesis2000
 patch=1,EE,0010e994,word,46011702
 patch=1,EE,0032f6ec,word,3f400000
 patch=1,EE,0032f850,word,3f2aaaaa
+
+
+[60 FPS]
+author=asasega
+comment=Patches the game to run at 60 FPS
+(EE Overclock should be set to 130%. Higher values are usable but may cause hanging).
```
refractionpcsx2, line 15: Can you keep the comment on the same line, it'll confuse the parser

J-McFc: Sure

refractionpcsx2: Thanks :)

PR #189, opened 2023-08-27
Added several widescreen patches and updated others patches

fox140cv:
all tested by me

### DIFF
--- a/patches/SLES-55380_8C913264.pnach
+++ b/patches/SLES-55380_8C913264.pnach
@@ -14,8 +14,9 @@ patch=1,EE,0043657c,word,3c024436 // 3c024420 2d char fix pos
 patch=1,EE,007638c0,word,3faaaaab // 3f800000 2d char fix fov
 
 
-[50 FPS]
+[60 FPS]
 author=PeterDelta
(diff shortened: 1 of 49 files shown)


patches/SLES-55380_8C913264.pnach
```diff
@@ -14,8 +14,9 @@ patch=1,EE,0043657c,word,3c024436 // 3c024420 2d char fix pos
 patch=1,EE,007638c0,word,3faaaaab // 3f800000 2d char fix fov
 
 
-[50 FPS]
+[60 FPS]
 author=PeterDelta
-comment=Unlocked at 50 FPS. Might need enable EE Overclock to be stable.
+comment=480p mode and FPS unlocked. Might need enable EE Overclock to be stable.
```
Mrlinkwii, line 19: can you make the 480p mode a separate  patch

Mrlinkwii: also if you want to make a 60 fps patch , make it a separate patch also to the 50 one

fox140cv: I had thought of doing it separately but I thought that when we played we would activate both and that's why I put it together, so as not to have so many patches. It looks better with progressive scan the truth. I have changed it

PR #191, opened 2023-08-30
Patches update

fox140cv:
Patch update, everything tested and working correctly

### DIFF
--- a/patches/SCES-51719_44A61C8F.pnach
+++ b/patches/SCES-51719_44A61C8F.pnach
@@ -31,3 +31,4 @@ patch=1,EE,E0030001,extended,00A57E7C
 patch=1,EE,206184D4,extended,00000051
 patch=1,EE,20618504,extended,00000240
 patch=1,EE,20618508,extended,000003c0
+patch=1,EE,20A57E70,extended,00000000
--- a/patches/SCES-51719_44A61C8F.pnach
+++ b/patches/SCES-51719_44A61C8F.pnach
@@ -18,8 +18,10 @@ patch=1,EE,204959B8,extended,4614D682 // mul.s f26, f26, f20
 patch=1,EE,204959BC,extended,08124B83 // j 00492E0C
 patch=1,EE,2044DF60,extended,00000000 // nop
 
-[No-Interlacing]
-gsinterlacemode=1
-author=Silent
-comment=480p mode in race. Set to Graphics "Deinterlacing: Adaptive (Top Field First)". Fix tremor at start game
-patch=1,EE,20A57E70,extended,00000001
+[Mode 1080i]
+author=PeterDelta
+comment=1080i mode in race. Disable it before entering in zone Replay Theatre. Disable all deinterlacing patches.
+patch=1,EE,E0030001,extended,00A57E7C
+patch=1,EE,206184D4,extended,00000051
+patch=1,EE,20618504,extended,00000240
+patch=1,EE,20618508,extended,000003c0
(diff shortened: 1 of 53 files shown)


refractionpcsx2:
Why did you remove a load of no-interlace modes?

refractionpcsx2:
Okay but not everybody wants 1080i, franky it looks almost as crap as normal interlaced mode, and you can't activate 480p mode in PAL versions, there's no option, that's why the no interlacing patch was there.

patches/SCES-51719_44A61C8F.pnach
```diff
@@ -18,8 +18,15 @@ patch=1,EE,204959B8,extended,4614D682 // mul.s f26, f26, f20
...
 
-[No-Interlacing]
-gsinterlacemode=1
+[Mode 480p]
 author=Silent
-comment=480p mode in race. Set to Graphics "Deinterlacing: Adaptive (Top Field First)". Fix tremor at start game
-patch=1,EE,20A57E70,extended,00000001
\ No newline at end of file
+comment=480p mode in race. Set to Graphics "Deinterlacing: Adaptive (Top Field First)". Incompatible with 1080i
+patch=1,EE,20A57E70,extended,00000001
```
refractionpcsx2, line 24: you missed the gsinterlacemode=1

refractionpcsx2: Also I don't think these patches are 480p but 240p

fox140cv: This is not interlaced, it only activates 480p and if we have it activated by default it would cause problems. That's why I select that option in graphics so that the screen does not shake at startup. If it were deinterlacing the screen wouldn't shake

fox140cv: > Also I don't think these patches are 480p but 240p

It may be, it will be forced, the resolution marks that in the emulator

refractionpcsx2: I didn't say it was interlace, but I'm pretty sure it doesn't activate 480p mode but 240p progressive mode (half resolution), which looks crap in native but good when upscaled.

Maybe @CookiePLMonster can clarify.

fox140cv: here it is published by him
https://forums.pcsx2.net/Thread-No-interlacing-codes?pid=620650#pid620650
// force progressive scan
patch=1,EE,20A57E70,extended,00000001

refractionpcsx2: yes, force progressive scan is not the same thing as 480p, 240p is probably what it's setting.

refractionpcsx2: just change it back to No-Interlacing, like it was before you removed it, all I asked is you restore the old patch.

fox140cv: Changed, I put that patch there myself so that it would start with that mode by default, and seeing that it is not deinterlacing, it seemed appropriate to leave it as an independent patch so that it would not open conflicts.

patches/SCES-54552_CBB4B383.pnach
```diff
@@ -17,4 +17,23 @@ patch=1,EE,00261FB4,word,3C020000 //3C023F80
 patch=1,EE,001D4E2C,word,24060000 //02C23021 Top
 patch=1,EE,001D4E7C,word,240201C0 //00451021 Bottom
 
-
+[50/60 FPS]
```
Mrlinkwii, line 20: is it 50 or 60 , it cant be both

refractionpcsx2:
Thanks :)

PR #192, opened 2023-09-01
Added Star Wars Episode III - Revenge of the Sith NTSC 60 FPS patch.

J-McFc:
I've played through the whole game with the patch and had no issues, although the main menu FMV is double speed but not game breaking of course.

### DIFF
--- a/patches/SLUS-21143_976EA156.pnach
+++ b/patches/SLUS-21143_976EA156.pnach
@@ -6,4 +6,7 @@ comment=Widescreen Hack
 patch=1,EE,004dbd2c,word,3c033fab
 patch=1,EE,0051ac44,word,3c023f20
 
-
+[60 FPS]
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
I would rather the FMV issue was fixed, before we merge it, maybe the patch could be made conditional on something else to switch when needed.

fox140cv:
This solves it, it is from the PAL version that also works here:

patch=1,EE,E0010002,extended,001B4344
patch=1,EE,201AF538,extended,00000001

I was going to add it along with other patches I'm testing, so we already have this one, try it and add it.

J-McFc:
Yep this works perfect for me, FMV plays at normal speed. I'll update it.

refractionpcsx2:
Awesome, thanks folks, good teamwork :)

PR #197, opened 2023-09-08
Update patches

fox140cv:
A few weeks ago I included patches deinterlacing in gt4 and Tourist Trophy with gsinterlacemode=1 that in reality only force progressive scanning, since when I included it the screen vibrated in the menus and when I removed it it worked perfectly, I come to correct that error resulting from my ignorance, now I know that only those intended for deinterlacing have to go there, that at least in these games they are not necessary with new versions of the emulator. With these changes they run perfectly with the emulator's default options without the need for any patch, the screen does not vibrate and looks correct.
I hope I expressed myself well. I've tested the rest and they work fine.

### DIFF
--- a/patches/SCES-50294_B590CE04.pnach
+++ b/patches/SCES-50294_B590CE04.pnach
@@ -3,4 +3,12 @@ description=Attempts to disable interlaced offset rendering.
 gsinterlacemode=1
 patch=1,EE,2025B4E8,extended,34020001
 
-
+[60 FPS]
+author=PeterDelta
+comment=Forces progressive scan and run at 60 fps
+patch=1,EE,2028F840,extended,24120052
+patch=1,EE,2035163C,extended,3C888889 //speed flags 
+patch=1,EE,20351F1C,extended,3C888889 //speed cars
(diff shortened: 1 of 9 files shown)


CookiePLMonster:
Do I understand correctly that this makes Gran Turismo 4 and Tourist Trophy PAL run at 60FPS, with the timer fixed up not to count too fast? Are you sure all the physics etc. remain consistent to how they are at 50FPS? Do keep in mind GT4 PAL came months after NTSC and if I recall specifically the developers specifically mentioned "reworking the physics to work correctly at 50FPS".

fox140cv:
You're right, it seemed that everything was working correctly, I did a few races with different cars and well, I just realized that it goes a little faster when returning to 50. I guess it was that 10% when increasing the fps and I didn't perceive it well...

CookiePLMonster:
> Should we eliminate them until we find a way to do it like in Gran Turismo 3?

I'd be inclined to agree with that yeah.

fox140cv:
Thanks for noticing, I have deleted it, I think it is better that way.

PR #203, opened 2023-09-11
Patches: Disable Castlevania: Lament of Innocence Widescreen Patch

RedPanda4552:
image
Patch does not AR adjust item icons in buy menus. Might be a good idea to disable until someone can work out how to fix it.


israpps:
on my archive, I have different cheats for this file.

Maybe you can scavenge something useful out of it?
https://github.com/PS2-Widescreen/OPL-Widescreen-Cheats/blob/main/CHT/SLUS_207.33.cht

refractionpcsx2:
Item Pos-X Fix does look promising, that could be related.

PR #206, opened 2023-09-20
Patches for Burnout 3 (60 FPS menus & crashes, always ask for 480p output, MPH to KPH)

ElTioRata:
Also added Aero_ as the widescreen fix author at the beginning.


fox140cv:
As a suggestion, since the patches that unlock 30fps are aiming for the same goal, why don't you combine them all into a single 60 fps patch?, except the progressive scan one

ElTioRata:
Not a bad idea but I'd keep 60 FPS crashes as a separate patch.

PR #208, opened 2023-09-23
Patch files for Tony Hawk's American Wasteland

ReveriePass:
Removes two filters that are problematic with hardware upscaling. Cheats by TechieSaru.


pgert:
Typo: "Removes Removes" -> "Removes".

refractionpcsx2:
Issue mentioned above has still not been corrected, please fix this so we can merge, thanks.

PR #212, opened 2023-10-02
No Interlace for Gran Turismo 3 update

xan1242:
Converted from asasega's code on the forums.

### DIFF
--- a/patches/PBPX-95503_8AA991B0.pnach
+++ b/patches/PBPX-95503_8AA991B0.pnach
@@ -1,0 +1,184 @@
+gametitle=Gran Turismo 3 A-Spec [NTSC-U] (PBPX-95503)
+
+[No-Interlacing]
+gsinterlacemode=1
+author=asasega
+patch=1,EE,2025A3C8,extended,34020001
+description=Attempts to disable interlaced offset rendering.
+// uncomment for the effect described below
+// comment=Press SELECT+L2 buttons to display the 480 buffer, press SELECT+R2 buttons to display the default buffers (the normal default mode)
+// patch=1,EE,D02EA812,extended,0000FEFE
+// patch=1,EE,204A6A20,extended,00000001
+// patch=1,EE,D02EA812,extended,0000FEFE
+// patch=1,EE,20250D84,extended,340494D3
+// patch=1,EE,D02EA812,extended,0000FDFE
+// patch=1,EE,204A6A20,extended,00000003
+// patch=1,EE,D02EA812,extended,0000FDFE
+// patch=1,EE,20250D84,extended,00832025
+
+[No-Interlacing-2]
+gsinterlacemode=1
+author=xan1242
+description=Experimental full-frame render hack for maximum sharpness in gameplay. WARNING: potentially breaks minor things.
+
+//
+// resolution stuff
+//
+patch=1,EE,00243204,short,3F80
+patch=1,EE,2024341C,short,01C0
+// this affects the actual res
+// field 0
+patch=1,EE,20243260,short,01C0
+// field 1 -- we nuke this one
+patch=1,EE,202432CC,short,0001
+patch=1,EE,202432F4,word,AC530008
+// disable field switching
+patch=1,EE,2025A3C8,extended,34020001
+
+patch=1,EE,2010148C,extended,24030000
+patch=1,EE,201014A4,extended,24020001
+// disable switchers
+patch=1,EE,20101490,extended,00000000
+patch=1,EE,201014B0,extended,00000000
+
+// force full frame
+patch=1,EE,204A6A20,extended,00000001
+
+// disable half-res rendering
+//
+patch=1,EE,20252A44,extended,00003021
+patch=1,EE,20252A70,extended,00003021
+patch=1,EE,20252ABC,extended,00001821
+patch=1,EE,20252AE4,extended,00002021
+patch=1,EE,20252B10,extended,00003021
+
+//
+// font stuff
+//
+patch=1,EE,201D1B54,short,01C0
+patch=1,EE,20180B6C,short,01C0
+patch=1,EE,20180CDC,short,01C0
+
+// halve 3D render res in menus to avoid bugs (Q29gt2k_menu3Car)
+// entrypoint
+patch=1,EE,201CAE7C,extended,0C0D2226
+patch=1,EE,201CAE98,extended,0C0D2226
+// detour to 0x00348898
+patch=1,EE,20348898,extended,03E03821
+patch=1,EE,2034889C,extended,0C091146
+patch=1,EE,203488A0,extended,00000000
+patch=1,EE,203488A4,extended,00E0F821 // MAX RES: 640x320, SAFE RES: 640x224
+patch=1,EE,203488A8,extended,24030280 // X res
+patch=1,EE,203488AC,extended,AC830008
+patch=1,EE,203488B0,extended,24030140 // Y res 
+patch=1,EE,203488B4,extended,AC83000C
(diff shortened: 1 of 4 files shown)


xan1242:
I've added a converted PAL version.

KingJackSkellington made a PR for his previously, but I've tested it and it didn't work, so I just did it myself.

xan1242:
Actually nevermind, you already had a PAL patch I see.... I will revert my PAL commit.

pgert:
About GT3 PAL:
Have you tested both hacks to see if there's any difference?

xan1242:
I hadn't tested the one from this repo but I did test my own and it does exactly what it's supposed to  - no jumpy picture, no interlace.

This one here seems to force 480p and 60Hz/FPS which sounds better for the user experience, but it's not accurate to call it a simple no interlace patch.

If you want to, we can split them into 2 separate patches, with mine being the "no-interlace" code and the other being the "60 FPS" code.

xan1242:
The one you have upstream appears to be made by junfurano yes, but yesterday I took [the code from this post](https://forums.pcsx2.net/Thread-No-interlacing-codes?pid=589919#pid589919) and manually converted it.

It appears they patch the same thing anyway, so we basically came down to the same conclusion. asasega's variant has the joker commands to control the modes, so I'll just extend it with that and it should be good.

xan1242:
How about this, is this any good?

fox140cv:
I tried that trick in the past on the US version and it removes the HUD, sun, smoke, etc... the image looks sharper but a lot of elements are sacrificed. I don't know if you've tried the port of the PAL version but I just tried it and when I press select+L2 the image turns black.

xan1242:
Well if it causes more problems we can just nuke the joker commands and trim it down only to no-interlace portion.

fox140cv:
Better, since if the 60fps progressive patch does not improve, we are going backwards
This code when pressing select+L2:
[No-Interlacing]
gsinterlacemode=1
author=asasega / junfurano / xan1242
comment=Press SELECT+L2 buttons to display the 480 buffer, press SELECT+R2 buttons to display the default buffers (the normal default mode)
description=Attempts to disable interlaced offset rendering.
patch=1,EE,2025B4E8,extended,34020001
patch=1,EE,D02EBE42,extended,0000FEFE
patch=1,EE,204A83E0,extended,00000001
patch=1,EE,D02EBE42,extended,0000FEFE
patch=1,EE,20251EA4,extended,340494D3
patch=1,EE,D02EBE42,extended,0000FDFE
patch=1,EE,204A83E0,extended,00000003
patch=1,EE,D02EBE42,extended,0000FDFE
patch=1,EE,20251EA4,extended,00832025

Gran Turismo 3 - A-Spec_SCES-50294_20231003213116

xan1242:
OK, nuking joker commands then... They cause more trouble than it's worth.

xan1242:
Actually uhh, I _may_ be cooking something entirely else... I got full field rendering together with the UI. Patching out half-height values seems to almost do the trick?

image

I'll poke around some more and I'll let you know the results...

xan1242:
OK, I have _something_ but I decided to not push it out yet.

I have a working full frame render:

image

However, the issues are as follows:
- Car rendering in menus is completely borked
image

- License replay window is borked
image

- NTSC-J version's HUD is squished at half res positions
image

- PAL version just flickers like crazy even with disabled field flipping

If you want, I can include the one for US 1.00 build as an experimental patch for now. Other versions will need further research.

fox140cv:
That's right, that's the bad thing about this trick, I was also doing tests in the past, especially in the PAL version and it's not possible, coming to the conclusion that the best thing is to force the progressive because it improves everything without losing anything.

xan1242:
You can't force progressive without halving the vertical res on this game. The game is hardcoded to render half the res on each field, and if you enable progressive, you force one field to render at a time. So in PAL's case, you'd go from 640x512 to 640x256 by enabling progressive.

The thing I made is coming together quite nicely, actually. I'll be sure to push it out once I fully debug it with PAL and NTSC-J as well.

xan1242:
Please test this out. I've written completely new codes. It forces the full frame resolution on one of the fields and disables the other, effectively giving progressive scan and full resolution during gameplay.

Current caveats:
- Menu car rendering is done in 640x320 (640x256 in PAL) resolution because it seems to run out of memory, therefore textures glitch out. You can test this by increasing it in the patch itself. This was done as a compromise to avoid buggy car textures in menus.
- Interlaced movies only show the first field, not the second, effectively halving all video resolution - this isn't something that is fixable easily. Maybe making progressive PSS video files and forcing progressive mode would fix this.
- PAL version has buggy fade screens in replays and during countdown.
- There may be some other minor visual bugs like that that I may have missed. I looked through every menu and tried to fix everything I could.

I'll look into making a code for the 1.1 US build next, since I don't see any here.

xan1242:
SS Route 5's sky is not happy... It seems to flicker in a very specific checkered fashion, or completely goes out. :(

I'll try to figure that out later.

xan1242:
OK I managed to halve the render res on the roulette cars, thereby fixing the buggy textures.

The trophy is hacked to just draw by disabling one of its transformation functions. I'm working on this now.

xan1242:
The trophy screens should now be fixed (as well as the data logger/analyzer screens).

The trophies normally render in 4 passes. Unfortunately, when specifically the championship trophy is being rendered, it doesn't draw on screen at all. In fact, it completely bugs out on software rendering, so this requires hardware renderers.

The multi-pass rendering is being only disabled when the championship trophy is on the screen. For regular trophies, it should still be multi-pass. (Except on PAL, this doesn't work at all, so it was disabled at all times)

PAL version is the one worst off here, unfortunately.

pgert:
The forum could give you better support for experimental hacks, just say'n.

xan1242:
Eh, it's fine. I've done testing with a few friends of mine and it's all good. Nobody (on NTSC) had any issues.

PAL version's problems are the only things left to address here.

I mean, if you don't like the "experimental" wording, I can remove it from the description now.

refractionpcsx2:
If the patches don't work correctly we won't merge them, regardless of your wording, we don't want broken patches

xan1242:
They're not broken as far as my testing shows.

But I can't guarantee it won't break something obscure.

The state of the PAL version is something I'm addressing at the moment myself.

And you would've known it's not broken if you tried the patch for yourself.

Either way, it's your loss if you don't merge. This is the only properly working patch to my knowledge.

refractionpcsx2:
You just told me PAL was broken

And saying "" you'd never had known it was broken if you'd not tried it* isn't exactly instilling me with confidence in what you're submitting.

xan1242:
Yes I did say PAL was broken.

I also said I'm working on it.

You can wait before you merge.

Where's the problem here?

I did not expect that you don't test yourselves to see if it at least works for 5-10 minutes before merging.

I'm sorry for the harsh tone, but I do feel patronized here for no good reason. Please, just let me finish working on this and merge it once it's done.

xan1242:
I'm cancelling this. The patches will stay private until further notice.

xan1242:
I've documented the entire process and stated that NTSC-U and J patches work and that they're practically finished. Nothing was being "sneaked in" and it was **very** clear what I was doing here. Every step was even commented on.

You have demonstrated clear inability to manage this properly as a maintainer, which includes reading, minimum testing and basic decency and respect towards someone making a contribution. This was worked on for days with great care.

I'm not discussing this matter any further. You reap what you sow and you're not getting any of my work, this included. I am done.

refractionpcsx2:
Good riddance.

TheUnpunished:
Just wanted to play GT3 and to see if there are new patches coming through. What a shame.

PR #216, opened 2023-10-12
Add 120 FPS patch for Burnout Revenge (U)

Scirvir:
120 FPS for Burnout Revenge:
- Compatible with existing 60 FPS patches for Crash mode. Applying those patches at the same time as this one only makes said gamemode run at 120 FPS aswell
- All audio is sped up. Not particularly annoying for game sounds but very much so for music
- No major gameplay problems except for colission against very steep ramps. Video attached as an example.
https://github.com/PCSX2/pcsx2_patches/assets/106109476/0e75d04e-c770-42dd-ba01-a5aff28b4209

Proof: https://youtu.be/VQRiaeKFrc8

### DIFF
--- a/patches/SLUS-21242_D224D348.pnach
+++ b/patches/SLUS-21242_D224D348.pnach
@@ -44,3 +44,12 @@ patch=1,EE,201125EC,extended,00108002
 author=SuperType1/remco
 description=Makes the Crashes & Crash Mode run in 60 FPS.
 patch=1,EE,20104B9C,extended,90850608
+
+[120 FPS]
+author=Scirvir
+description=Requires setting PCSX2 framerate to 120FPS in Properties>Emulation>Speed Controls.
+patch=1,EE,10103238,extended,3c104205
+patch=1,EE,10104a30,extended,3c104205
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Sorry, I'm not really happy with 120fps patches becoming a thing, this is going to be dangerous for the actual emulation, the console was never designed to run beyond 60.

So It's a no from me, sorry.

TheUnpunished:
> the console was never designed to run beyond 60.
The console was never meant to run games at higher resolution, either.

refractionpcsx2:
yes but upscaling doesn't cause timing problems, as you even outlined in your original post (that list of broken things is way too long for me to be happy with the patch, even if that was a 60fps patch)

TheUnpunished:
Upscaling causes problems with effects. Bloom in particular, even though there are workarounds.
And I am not an original uploader.

refractionpcsx2:
yeah but you can work around it and they aren't game breaking, the 120fps changes are game breaking and have much bigger impacts internally than upscaling.

Apologies in the mistaken identity, I wasn't expecting somebody else to jump in to the defense.

PR #217, opened 2023-10-15
add 60fps patches and updates

fox140cv:
All have been tested by me and members of the forum working without problems


refractionpcsx2:
Thanks for helping test and bring these over :)

PR #218, opened 2023-10-17
Crash Twinsanity (SLUS-20909) Widescreen Patch

ReveriePass:
A patch for Crash Twinsanity (B318AA3C, 8CFAB4EA) that always enables widescreen.


refractionpcsx2:
Can you name the section header according to the pnach 2.0 documentation and add the aspect ratio setting, this just makes it work nicely for the user.

https://forums.pcsx2.net/Thread-Sticky-Important-Patching-Notes-1-7-4546-Pnach-2-0

PR #221, opened 2023-10-17
Patches: Various additions and improvements

CRASHARKI:
Changes:

- Add 50/60 FPS patches to The Simpsons Game
- Fix Max Payne patches to use correct CRCs
- Add Widescreen and 60 FPS patches to Shrek 2 + correct CRCs

All of the patches have been tested for all versions of the games, and relevant comments such as activating widescreen in-game or the need for EE Overclock have been added.

### DIFF
--- a/patches/SLES-50326_52E0597D.pnach
+++ b/patches/SLES-50326_52E0597D.pnach
@@ -1,4 +1,5 @@
 gametitle=Max Payne [PAL-M4] (SLES_503.26)
+//This pnach corresponds to the main CRC of the game, but it's just meant to activate SLES-50326_C669B63C.pnach
 
 [Widescreen 16:9]
 gsaspectratio=16:9
--- a/patches/SLES-50326_52E0597D.pnach
+++ b/patches/SLES-50326_52E0597D.pnach
@@ -2,15 +2,5 @@ gametitle=Max Payne [PAL-M4] (SLES_503.26)
 
 [Widescreen 16:9]
 gsaspectratio=16:9
-comment=Widescreen Hack by El_Patas
-
-//Gameplay 16:9
-//ELF file is called "MAIN_P.RUN"
-patch=1,EE,0050eae0,word,3c013f40 //00000000 (Increases hor. axis)
-patch=1,EE,0050eae8,word,4481f000 //00000000
-patch=1,EE,0050eaec,word,0c04821c //00000000
-patch=1,EE,0050eaf0,word,00000000 //0c04821c
-patch=1,EE,0050eaf4,word,461e0003 //00000000
-patch=1,EE,0050eb04,word,461e0082 //c68201f8
-
-
+author=El_Patas
+comment=Widescreen Hack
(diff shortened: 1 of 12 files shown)


patches/SLES-50326_52E0597D.pnach
```diff
@@ -2,15 +2,5 @@ gametitle=Max Payne [PAL-M4] (SLES_503.26)
...
-patch=1,EE,0050eae0,word,3c013f40 //00000000 (Increases hor. axis)
-patch=1,EE,0050eae8,word,4481f000 //00000000
-patch=1,EE,0050eaec,word,0c04821c //00000000
-patch=1,EE,0050eaf0,word,00000000 //0c04821c
-patch=1,EE,0050eaf4,word,461e0003 //00000000
-patch=1,EE,0050eb04,word,461e0082 //c68201f8
-
-
+author=El_Patas
+comment=Widescreen Hack
```
refractionpcsx2, line 6: Where did the patch go?

patches/SLUS-20230_BEB4577E.pnach
```diff
@@ -2,15 +2,5 @@ gametitle=Max Payne (NTSC-U)
...
-patch=1,EE,0050e030,word,3c013f40 // 00000000 hor fov
-patch=1,EE,0050e038,word,4481f000 // 00000000
-patch=1,EE,0050e03c,word,0c04821c // 00000000
-patch=1,EE,0050e040,word,00000000 // 0c04821c
-patch=1,EE,0050e044,word,461e0003 // 00000000
-patch=1,EE,0050e054,word,461e0082 // c68201f8
-
-
+author=ElHecht
+comment=Widescreen Hack
```
refractionpcsx2, line 6: Where did the patch go?

patches/SLES-50326_52E0597D.pnach
CRASHARKI: It's one of those situations where the main ELF isn't the one to apply the patch to, but it's required for the game to show the patches and be able to activate them. It is better explained here: https://github.com/PCSX2/pcsx2/issues/9243#issuecomment-1635168588
More specific: 
> I see the point about it being an edge case with a loader ELF - in this case, **I'd suggest duplicating the pnach file for both CRCs, but leaving the codes empty for the loader**.
> 
> We will **not** be restoring the old behaviour. It was broken from the beginning, and caused crashes/broken games **when patches were applying to the wrong ELF**.
I've tested it and this way it shows the patches on the game settings and applies the chosen patches on the right ELFs.

patches/SLUS-20230_BEB4577E.pnach
CRASHARKI: Same as the other one.

patches/SLES-50326_52E0597D.pnach
CRASHARKI: Done

refractionpcsx2:
Thanks :)

PR #224, opened 2023-10-27
Add Spyro default TV mode (SLUS 20315)

Muzzarino:
Sets, on boot the default screen mode for Spyro: Enter the Dragonfly to be set widescreen instead of standard.

### DIFF
--- a/patches/SLUS-20315_4B8E0DE8
+++ b/patches/SLUS-20315_4B8E0DE8
@@ -1,4 +1,5 @@
 gametitle=Spyro: Enter the Dragonfly (SLUS_20315)
 
 [Default Widescreen mode]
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
If you're going to force widescreen can you add the aspectratio parameter too please https://forums.pcsx2.net/Thread-Sticky-Important-Patching-Notes-1-7-4546-Pnach-2-0

Muzzarino:
> If you're going to force widescreen can you add the gsaspectratio parameter too please https://forums.pcsx2.net/Thread-Sticky-Important-Patching-Notes-1-7-4546-Pnach-2-0

Added.

refractionpcsx2:
cool, thanks :)

PR #231, opened 2023-11-03
Restore unjustly removed twinsanity widescreen patches

PrgmrAtHrt:
I am re-submitting my (TechieSaru's) patches that I strongly believe were replaced with malicious intent for the following three reasons:

1. The uploader mentioned that their update was a "fix" and delibrately chose not to specify what it was "fixing"
2. BOTH patches were replaced
3. I was never directly informed that there was an issue with my patch

The patch I originally created was by request for a friend (Reverie). It was for v2.00 of crash twinsanity and my friend and myself verified that it worked fine. 

Upon futher investigation, I figured out that the "fix" that the uploader was, again, being deceitful about, is due to my friend making an erroneus assumption that my hack worked for both versions of Twinsanity.

I have taken the opportunity to correct this error with my patch.

### DIFF
--- a/SLUS-20909_8CFAB4EA.pnach
+++ b/SLUS-20909_8CFAB4EA.pnach
@@ -1,0 +1,13 @@
+gametitle=Crash Twinsanity (v2.00) (SLUS-20909)
+
+[Widescreen 16:9]
+author=TechieSaru
+gsaspectratio=16:9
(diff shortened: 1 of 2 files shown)


PrgmrAtHrt:
Bad request... Reuploading.

PR #232, opened 2023-11-03
Restore unjustly removed twinsanity widescreen patches

PrgmrAtHrt:
I am re-submitting my (TechieSaru's) patches that I strongly believe were replaced with malicious intent for the following three reasons:

The uploader mentioned that their update was a "fix" and delibrately chose not to specify what it was "fixing"
BOTH patches were replaced
I was never directly informed that there was an issue with my patch
The patch I originally created was by request for a friend (Reverie). It was for v2.00 of crash twinsanity and my friend and myself verified that it worked fine.

Upon futher investigation, I figured out that the "fix" that the uploader was, again, being deceitful about, is due to my friend making an erroneus assumption that my hack worked for both versions of Twinsanity.

I have taken the opportunity to correct this error with my patch.

### DIFF
--- a/patches/SLUS-20909_8CFAB4EA.pnach
+++ b/patches/SLUS-20909_8CFAB4EA.pnach
@@ -1,8 +1,13 @@
 gametitle=Crash Twinsanity (v2.00) (SLUS-20909)
 
 [Widescreen 16:9]
+author=TechieSaru
 gsaspectratio=16:9
-author=CRASHARKI
-comment=Run the game at 16:9 Widescreen Aspect Ratio from the start.
-patch=1,EE,2030A744,word,00000001 //00000000
-patch=1,EE,20BE1DE4,word,00000001 //00000000
+patch=1,EE,20166E64,extended,24060001
+patch=1,EE,201798AC,extended,24020001
+patch=1,EE,2019B270,extended,24030001
+patch=1,EE,2019FE7C,extended,24020001
+patch=1,EE,2019FF0C,extended,24020001
+patch=1,EE,201A04C0,extended,24020001
+patch=1,EE,201A6E74,extended,24030001
+patch=1,EE,202AE764,extended,24020001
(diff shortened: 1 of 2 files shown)


refractionpcsx2:
@CRASHARKI can you explain yourself, please.

CRASHARKI:
Of course @refractionpcsx2. I created the patches I submitted months ago, back then I didn't know that patches that activate the widescreen for games that include it natively were being added so I didn't submit them. When I saw that @PrgmrAtHrt 's submission for this very game had been accepted I prepared to submit the other missing versions since I had done patches for all four versions of the game. I had submitted back then my patches to a server dedicated to the game, and never had a complaint about it, but when @PrgmrAtHrt 's patches were officially added people started having issues, and when I tested it, so did I. Since my patches worked fine and were less invasive (changing 2 values is safer than changing 8) I replaced both versions just to be safe.

I never did anything with malicious intent or by being deceitful, ironically you're now doing the same thing you're accusing me of by replacing not only the version that supposedly worked, but also the patch I made that actually worked with your own. @PrgmrAtHrt I would appreciate if in the future you refrain from calling people malicious or deceitful, give people the benefit of the doubt, after all, we all want the same thing, to make PCSX2 as good as possible.

I don't care which patch stays as long as it works as intended, but if it's yours, could you add the description to the patch so that people understand what it does?

Sorry for the long text, but I wanted to be as clear as possible.

refractionpcsx2:
Thank you for the explanation, I agree it wasn't fair to say it was malicious, we all just want working widescreen patches, I'm not precious about who's it is, I just want the best one, and to be honest a 2 line patch is better and looks less janky (from an outside perspective), if it does the job correctly, then sorry @PrgmrAtHrt , I'd rather keep that.

PrgmrAtHrt:
I updated my patches with comments. Mine is an assembly hack that forces the game into widescreen. The game reads the widescreen flag in eight different places in its code.

refractionpcsx2:
yes, and the existing code just changes the flag, from what I'm aware, instead of messing around finding anywhere it reads it

CookiePLMonster:
Do note that the write to `00BE1DE4` looks like a write to dynamically allocated memory - is that the case here? If so, the "simpler" patch may be less reliable.

refractionpcsx2:
Alright, I'll roll with this, but I don't wanna see this again, unless there's a *very* good reason everyone can agree on.

PR #238, opened 2023-11-04
Fix my mistakenly swapped Crash Twinsanity patches

PrgmrAtHrt:
I somehow ended up getting the two patches swapped with each other and had the wrong code in each patch file. I apologize for this mistake. I have verified that there are no issues.


refractionpcsx2:
No, I'm not doing this anymore, the other patch stays and is much simpler.

PR #239, opened 2023-11-08
Add more 60fps patches

fox140cv:
All tested, work without problem


refractionpcsx2:
@fox140cv just to let you know, I now no longer trust when you say `all tested and work without problem`

We just tested the killzone patch and not only do all the FMVs (including the main menu) run at double speed, but also the planes that fly over in the beginning of the game break and hang in mid air.

PR #241, opened 2023-11-09
Patch: True Analogs mode for AC2 AC2 - Another Age AC3 and AC - Silent Line

JordanTheToaster:
True analog mode by VanLaser.

https://github.com/VanLaser/analog-controls-for-armored-core-ps2


refractionpcsx2:
Sorry I don't feel a patch this long should be in our repo for changing game behaviour for convenience in this way.

I know you used the R&C 3 widescreen patch as justification, however I don't like that either, and from the looks of it, if we can make dynamic patches work with pnach files, that crazy amount of lines can be come like, 16, as most of it is just checking where it is in the code.  Not to mention patching dynamically changing code like that is *very* scary.

Anyway, it's a no from me, if somebody wants this they should patch their ISO.

PR #246, opened 2023-11-12
Maintain patches for Armored Core games

Berylskid:
Overhauled Widescreen patches and No-interlace patches of these games:
- Armored Core - Nexus
- Armored Core - Nine Breaker
- Armored Core - Formula Front

Note: Only for NTSC-J and NTSC-U.


AmyRoxwell:
Tested the NTSC-U patches for armored core nexus, they work great!!
Screenshot from 2023-11-15 14-56-19
Screenshot from 2023-11-15 15-00-43

PR #247, opened 2023-11-12
Add patches 60 fps

fox140cv:
Tested and working correctly


patches/SCES-50878_F48F994A.pnach
```diff
@@ -11,12 +11,3 @@ patch=1,EE,002015d8,word,44810000  // 00000000
...
 patch=1,EE,001f7028,word,3c013f40  // 3c013f80 partial HUD fix
-
-
-[No-Interlacing]
-gsinterlacemode=1
-author=felixthecat1970
-comment=Autoboot progressive scan mode
-patch=0,EE,001E2254,extended,24020002
-patch=0,EE,0022B138,extended,24050006
-patch=0,EE,001EDC24,extended,24020009
```
refractionpcsx2, line 22: Was there something wrong with this?

fox140cv: When an fmv is opened it goes back to PAL and does not change to progressive, as this patch causes the interlacing to be in "none" the image shakes in PAL.
Since the Hz selector (including progressive) is at the beginning, it is best to take a second to choose it since you cannot choose Hz in the game and with the patch active it skips this selector.

fox140cv: tekken 4 with patch

I attach screenshots to illustrate the explanation, you can clearly see how it starts well in progressive and when the intro comes out it changes to PAL and remains. If you skip the intro by pressing start on the first NAMCO image, it plays fine in progressive, but if you start watching some FMV in gallery or story mode, when a video comes out it goes back to PAL and stays.

tekken 4 without patch

The only way to select the Hz and the progressive is at the beginning with this screen that with the patch enabled is skipped

PR #251, opened 2023-11-20
Silent Hill 2/3 60 Patches

MrWrightyLC:
Silent Hill 2 NTSC SLUS-20228_FE06A030
Silent Hill 3 NTSC SLUS-20622_2498951B


MrWrightyLC:
Both games finished from start to finish

PR #252, opened 2023-11-21
Patches: 60 fps and widescreen

PeterDelta:
I add 60fps and widescreen patches, all tested and working without problem.

007 - Everything or Nothing_SLES-52005_20231122180117
Old patch, doesn't seem to work with new versions of PCSX2
007 - Everything or Nothing_SLES-52005_20231122180406
New patch

### DIFF
--- a/patches/SLES-50276_CC76CD02.pnach
+++ b/patches/SLES-50276_CC76CD02.pnach
@@ -1,0 +1,6 @@
+gametitle=Gift (PAL-E) SLES-50276 CC76CD02
+
+[50 FPS]
+author=PeterDelta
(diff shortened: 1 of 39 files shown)


pgert:
The James Bond Widescreen hack:
Perhaps the NTSC-J disc (SLPM-65446) and the PAL-M5 disc (SLES-52005) has the same CRC.
So, unless you can confirm that the hack by nemesis2000 & Arapapa doesn't work for the NTSC-J disc, then better would be to have to files, one for the NTSC-J disc (SLPM-65446) with this hack, and another for the PAL-M5 disc (SLES-52005) with your hack.

PR #253, opened 2023-11-23
MotorStorm and Hulk 60fps Patches

FrankyBuster:
Adds tested 60fps patches for two games


patches/SCES-55573_3C967DB9.pnach
fox140cv, line 1: Hi, this is well added. You are not really changing anything here, just the author and the name of the patch.
The author is El_Patas porting the NTSC version of asasega
https://forums.pcsx2.net/Thread-60-fps-codes?pid=554224#pid554224
patch=1,EE,2039B4F8,extended,00000000 and patch=1,EE,0039B4F8,word,00000000 is the same

FrankyBuster: Sorry, i noticed just after making the pull request and i didn't know how to undo it.

PR #259, opened 2023-12-08
Midnight club 2 and 3 widescreen correction

PeterDelta:
Based on the 4:3 ratio, I have corrected it and would say it is accurate. With the previous value it is a little narrower. I attached images

MC2 4
MC2 16

MC3 4
MC3 16


Mrlinkwii:
have the patches been tested ?

refractionpcsx2:
Thanks :)

PR #262, opened 2023-12-13
Update 60 fps: Sonic heroes and unleashed

PeterDelta:
Sonic Heroes: You can choose between 50 and 60hz without altering anything in the game if 50hz is selected by mistake, this way it is more faithful. Selecting 60 Hz automatically switches to progressive mode

Sonic Unleashed: Fixed aspect ratio as shown in the images.
What I have marked with a red arrow is from the panoramic patch, when enabled it cannot be hidden because it is part of the map. I haven't been able to fix it but it doesn't affect the gameplay.

Sonic unleashed

Sonic Unleashed_SLES-55380_20231216025210

Update: In the end I found the direction of widescreen, it does not require a rendering fix and works very well, plus the HUD looks better this way without showing that part of the map that I marked in red.
___________________________________________________________________________________________________________________________________

Sonic Gems Collection: The widescreen patch does not correspond to this crc. It corresponds to the Japanese version and is included in this repository SLPM-66074_E0068D0A. It seems that it is for the Sonic R but in the PAL version it does not work and also the Sonic R would have a different crc than this one.

### DIFF
--- a/patches/SLES-51950_B7CF071A.pnach
+++ b/patches/SLES-51950_B7CF071A.pnach
@@ -1,12 +1,10 @@
-gametitle=Sonic Heroes (PAL-M5) (SLES-51950)
+gametitle=Sonic Heroes (PAL-M5) (SLES-51950) B7CF071A
 
 [Widescreen 16:9]
 gsaspectratio=16:9
 author=ElHecht
-comment=Widescreen Hack ported from NTSC hack by nemesis2000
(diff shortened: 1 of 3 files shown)


Mrlinkwii:
may i ask why SLES-53350_4A198252 was deleted ?

Mrlinkwii (approved):

PR #263, opened 2023-12-17
Project Zero - Remove Post Processing patch update

MrWrightyLC:
this patch used to be mixed with [No interlacing]
so I decided to separate these patches, as one of them is an attempt against the atmosphere of the game.
now the user will be able to disable/enable as desired

without patch
Project Zero_SLES-50821_20231217171825
Project Zero_SLES-50821_20231217171955
with patch
Project Zero_SLES-50821_20231217172109
Project Zero_SLES-50821_20231217172028


refractionpcsx2:
Thanks for separating that out, dunno how that snuck in....

PR #264, opened 2023-12-19
Burnout Revenge/Dominator Patches Update

MrWrightyLC:
### Burnout Revenge (U)(SLUS-21242)
**[Progressive Scan] always ask for 480p**
**[MPH to KPH] Miles per hour to Kilometres per hour speedometer**

### Burnout Dominator (U)(SLUS-21596)
**[Widescreen 16:9 and 21:9] new entire patch by SuperType1 (https://www.youtube.com/watch?v=fXY97XEIQQ8)**
**[60 FPS Menus and Crashes] 60fps menus and crash cam**

PR #267, opened 2023-12-21
Burnout Revenge - Always ask 480p, MPH to KPH patches

ElTioRata:
And I also fixed the misspelled author's name on Burnout 3's pnach (didn't noticed until now)


ElTioRata:
closed in favor of #264 looks like the author's has actually two aliases

PR #265, opened 2023-12-20
Move Test Drive - Eve of Destruction Patches to "patches" Folder

Virjoinga:
I mistakenly put it in the main folder instead of the patches folder

PR #248, opened 2023-11-14
Test Drive - Eve of Destruction Patches

Virjoinga:
Add patches for changing aspect ratio and disabling 16:9 edge borders (main menu and paint shop).

PR #272, opened 2023-12-24
Remove wrong pnach

Berylskid:
Just removed `SLUS-20910_5D0244D3.pnach` from wrong directory.


Daniel-McCarthy:
Just for added context for convenience:
This PR #248 added this file to the root folder accidentally, and this PR #265 copied the patch into `/patches/` folder but didn't delete the copy left in the root folder. 

Figured I'd add this as a note since I had checked into why this was here and if there was a version where it was supposed to be as well. LGTM!

Daniel-McCarthy (approved):
LGTM
Verified that the file being removed also exists in `/patches/` and that the file contents are identical so nothing is lost. 👍

PR #273, opened 2023-12-26
Remove no-interlaced patches

fox140cv:
With the new revisions, pcsx2 does an excellent job with interlacing and there are really very few games that require this.
As seen in the images, the interlacing patch prevents the letters from displaying correctly and the sawtooth becomes more noticeable in the graphics, even if we increase the resolution.
On the left side interlaced patch enabled and on the right interlaced patch disabled

Baldurs gate 1
Baldur's Gate - Dark Alliance
Baldurs gate 2
Baldur's Gate - Dark Alliance 2
champions return to arm
Champions - Return to Arms
extermination
Extermination
ghost vibration
Ghost Vibration

### DIFF
--- a/patches/SCES-50240_68707E85.pnach
+++ b/patches/SCES-50240_68707E85.pnach
@@ -8,12 +8,3 @@ comment=Widescreen Hack
 //Gameplay 16:9
 patch=1,EE,001D3158,word,3C023F19 //3C023F4C (Increases hor. axis)
 patch=1,EE,001D315C,word,3442AAAB //3442CCCD
-
-
-[No-Interlacing]
-description=Attempts to disable interlaced offset rendering.
-gsinterlacemode=1
-patch=1,EE,00101614,word,00000000
-patch=1,EE,0010187c,word,00000000
-
-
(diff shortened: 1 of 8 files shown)


refractionpcsx2:
Are you sure? I'm pretty sure when you upscale, the interlacing kinda goes to shit in the UI font

fox140cv:
> Are you sure? I'm pretty sure when you upscale, the interlacing kinda goes to shit in the UI font

Maybe that happened before but certainly not now, from what I see.
In Extermination and Ghost Vibration there is hardly any difference in the font, if you notice the sawtooth of some graphics that are accentuated with the patch enabled.
They have an upscale of x3 and unless I have overlooked some specific point in the UI that you mention, the conclusion I draw is that there is no improvement, it only makes it worse.

bald 1

bald 2

champion

extermination

Ghost Vibration

refractionpcsx2:
Which picture is which in your screenshots? apart from the last one none of them are labelled

5x native looks kinda terrible tbh :P 

image

refractionpcsx2:
But no, you're right, the no-interlace patch is demonstratably worse.

I'll accept this :)

PR #277, opened 2023-12-29
Patches: Adding 60 fps

fox140cv:
I add tested 60 fps patches

### DIFF
--- a/patches/SLUS-20946_2C6BE434.pnach
+++ b/patches/SLUS-20946_2C6BE434.pnach
@@ -1,10 +1,9 @@
-gametitle=Grand Theft Auto: San Andreas (SLUS-20946GH) / Ver 3.00
+gametitle=Grand Theft Auto: San Andreas (SLUS-20946GH) 2C6BE434 / Ver 3.00
 
 [Widescreen 16:9]
 gsaspectratio=16:9
-comment=Widescreen fix by nemesis2000 (pnach by nemesis2000)
-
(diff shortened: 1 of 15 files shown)


patches/SLUS-20946_2C6BE434.pnach
```diff
@@ -16,6 +15,10 @@ patch=1,EE,00242DB4,word,0C044C32
...
 [Remove Radiosity Filter]
 description=Removes the radiosity filter which causes a ghosting effect on the people and environment.
-
 //Remove Radiosity Filter
 patch=1,EE,2051A6D8,extended,00000000
+
+[60 FPS]
+author=someother1ne
+description=Unlocked at 60 FPS. Might need enable 130% EE Overclock to be stable.
+patch=1,EE,0066804C,word,10000001
```
pgert, line 24: patch=1,EE,0066804C,word,10000001
patch=1,EE,D066804C,word,10000001
https://forums.pcsx2.net/Thread-60-fps-codes?pid=637603#pid637603

fox140cv: That patch makes no sense, as it is published the second line does nothing and that is why I did not include it. The game runs without errors.
I have checked it and at the beginning of the game it has one value and from the loading screen it has another. It has to be like this, I have already corrected it:
patch=1,EE,D066804C,extended,00000002
patch=1,EE,0066804C,extended,10000001

PR #281, opened 2023-12-31
Adding widescreen and fps patches

PeterDelta:
The patches are tested and work correctly.

Changes in Rayman 2 - Revolution. The previous patch maintains 50fps at all times and makes some scenes play twice as fast. Some examples of these scenes:

rayman old
Previous patch running at double speed

rayman new
New patch fixed

rayman new usa

It works the same in ntsc-u version, in gameplay it changes automatically


Mrlinkwii:
have these been tested to make sure theirs no crashes etc ?

PR #286, opened 2024-01-05
Patches: Clean up Persona 3 FES NTSC-K patch

JordanTheToaster:
Removes forced 480p 60 fps mode that causes double speed and party control mod that was not in it's own section.


pgert:
Shouldn't "comment=Controllable party members by TGE (NTSC-K by Arapapa)" be removed as well?

Mrlinkwii (changes requested):
you missed "comment=Controllable party members by TGE (NTSC-K by Arapapa)" other wise good

refractionpcsx2:
probably should have moved the controllable party member patch in to its own header, rather than just deleting it.

PR #287, opened 2024-01-05
WS: make the Project Zero WS patch less likely to crash

Mrlinkwii:


patches/SLES-50821_22E91837.pnach
```diff
@@ -11,19 +11,14 @@ comment=Widescreen Hack
...
-; ==========
-; Due to missing renderingfixes, black stripes can show up on ghosts, spirits and photos.
-; If using GSdx in HW-mode, this can be fixed by having "Enable HW Hacks" set in the GSdx configuration dialog,
-;  and having "Align Sprite" set and "Sprite" set to Half in "Advanced Settings and Hacks".
-; ==========
 
-; ==========
+// ==========
 // 16:9 support.
 ; ==========
```
pgert, line 17: What about this one?

pgert:
Can ";" instead of "//" cause crash?
Thats news to me.

patches/SLES-50821_22E91837.pnach
Mrlinkwii: ah my bad ill fix that , thanks for catching

refractionpcsx2:
; is not a valid comment delimiter for the patch system

PR #289, opened 2024-01-06
Update SLES-51011_989192FE.pnach

PeterDelta:
Add conditional to be able to select a language at the beginning and thus keep the patch enabled from the pcsx2 patches section.

Knight Rider - The Game_SLES-51011_20240106213536


Mrlinkwii (approved):
assumed tested

PR #292, opened 2024-01-09
Adding patches 50/60fps

PeterDelta:
I add 50/60 fps patches, tested without errors.
Added widescreen patch for Disney Piglet - The Big Game (PAL-S)

* Dragon Quest VIII: fixed a rare bug in Angelo's ability "flaming lunge" when he used a specific type of sword.
* True Crime - New York City: Unlocked indoors only, outdoors AI is buggy
* Ridge Racer V: Added interlacing patch, letters now look sharp
* F1 Career Challenge: Add patch 50 fps and native widescreen,  fix circuit map aspect ratio, rear view mirrors and reflections

F1 Career Challenge 4 3
4:3 original

F1 Career Challenge old patch
Old patch

F1 Career Challenge new
New patch
I run it in software mode because in hardware mode it has graphical errors in the backgrounds of the circuits (it's a matter of the emulator, not the patch).

### DIFF
--- a/patches/SLES-51584_2C1173B0.pnach
+++ b/patches/SLES-51584_2C1173B0.pnach
@@ -2,11 +2,11 @@ gametitle=F1 Career Challenge (PAL & NTSC-U)(SLUS-20693 & SLES-51584)
 
 [Widescreen 16:9]
 gsaspectratio=16:9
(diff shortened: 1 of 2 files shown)


Mrlinkwii (approved):
assumed tested

PR #294, opened 2024-01-12
 [Tokyo Xtreme Racer 3] - Pnach file for the patched iso of the game

MarcoEstevez:
> When Tokyo Xtreme Racer 3 released in America, the localization team made one big mistake. They converted all the Credits from Japanese Yen to American Dollars by dividing the value by 100. You start with 40,000 credits instead of 4,000,000 million credits and so on.
> 
> However while making that change work they never changed the requirement for two wanderers. One of them (Exotic Butterfly)was left at 20 million credits (more credits then you’ll earn in an entire run of the game), and the other (Whirlwind Fanfare) was left at 100 million credits (more credits than you can ever earn as there’s a cap on your credits at 99,999,999. These are the same values as they are in the Japanese release.

recently a "fix" has been created in other to apply it to game so that no cheat codes are needed anymore

[https://www.romhacking.net/hacks/8339/](https://www.romhacking.net/hacks/8339/)

However widescreen and No-Interlacing patches are missing for the patched iso

This PR will address that
image

image


MrWrightyLC:
Correct me if I'm wrong but I think this PR  https://github.com/PCSX2/pcsx2_patches/pull/280 fixes that problem so it wouldn't be necessary to use a patched rom.
Oh and the name of your patch is wrong it should include the serial number

MarcoEstevez:
already addressed by https://github.com/PCSX2/pcsx2_patches/pull/280

pgert:
It might not be necessary for anyone to use a patched rom anymore, but some might still do it.
If you correct the pnach-name into "SLUS-20831_0F9348FF" then there shouldn't be any problems.

MarcoEstevez:
Fixed fileName to follow naming convention, feel free to merge if you want, as @pgert said, this only applies for those who have the patched rom

Mrlinkwii (approved):
assumed tested

MarcoEstevez:
Yes, I did tested successfully

PR #302, opened 2024-01-30
New WS patch for Dirge of Cerberus - Final Fantasy VII

### LINKED ISSUE #295: Final Fantasy 7 Dirge of Cerberus WS Patch causes black void in sky
For SLUS-251419 at least, there seems to be problems with the clip plain in the WS patch, which causes a black void to appear in the middle of the screen in the distance. (Also the UI isn't scaled correctly, it's all around a bad patch) 

image

We will disable the patch or now until somebody can fix it.

Gabominated:
Older patch have render issues as refraction exposed at issues section, I made a new patch that fix that missing of render and correctly scaling, so no more black hole at the sky

Dirge of Cerberus - Final Fantasy VII_SLUS-21419_20240128174517
Dirge of Cerberus - Final Fantasy VII_SLUS-21419_20240128183138

Fixes #295

### DIFF
--- a/patches/SLUS-21419_44A5FA15.pnach
+++ b/patches/SLUS-21419_44A5FA15.pnach
@@ -1,9 +1,10 @@
-//[Widescreen 16:9]
-//gsaspectratio=16:9
-//comment=Dirge of Cerberus - Final Fantasy VII Widescreen Hack (16:9) (NTSC-U)
+gametitle=Dirge of Cerberus - Final Fantasy VII SLUS-21419 44A5FA15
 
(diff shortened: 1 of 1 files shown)


Mrlinkwii (approved):
assumed tested

Gabominated:
> assumed tested

Yeah I was testing a lot and I don't see any black hole at the sky

refractionpcsx2:
Thanks for fixing :)

Gabominated:
> Thanks for fixing :)

It's my pleasure, this is a great game and we most enjoy it with more field of view

Issue #303, opened 2024-01-30
Invalid Widescreen patch line for Spyro: A Hero's Tail


Daniel-McCarthy:
### Description:

For "Spyro: A Hero's Tail" [SLUS-20884_94C56923.pnach](https://github.com/PCSX2/pcsx2_patches/blob/main/patches/SLUS-20884_94C56923.pnach#L10) and [SLES-52569_6E51213C.pnach](https://github.com/PCSX2/pcsx2_patches/blob/main/patches/SLES-52569_6E51213C.pnach#L10), the widescreen patches have a broken patch line that will not be loaded due to using `patch=3` which is not a valid option [as per Patch.cpp](https://github.com/PCSX2/pcsx2/blob/master/pcsx2/Patch.cpp#L53).

The behavior is simply that the line gets ignored as if it doesn't exist.
Since the 3 patch lines use this field as a "patch number", they put each line as 1,2,3, so all 3 lines need to be updated to 0, 1, or 2 depending on what is most appropriate for this patch.
```
//Widescreen + Extra Perspective
patch=1,EE,003B28C8,extended,3C013F10
patch=2,EE,003B28DC,extended,3C013F10
patch=3,EE,003B2F08,extended,3C013EC0
```

As I don't have this game, I can't test the impact of the change so am reporting instead of submitting a PR to fix it.

Mrlinkwii:
i managed to test the us version  , with the  `patch=3,EE,003B2F08,extended,3C013EC0 ` removed ,  i saw no difference and nothing broke 
![image](https://github.com/PCSX2/pcsx2_patches/assets/17685135/5ed27d40-6bab-411d-bf08-eb213612a6d8)

PR #304, opened 2024-01-31
Adds Action to validate patches for new PRs

Daniel-McCarthy:
# Description:

Adds a CI action that validates all patches in the repo based on criteria (detailed below) when a PR is created or updated with new commits. It will report pass or failure based on if any errors were detected with the patches. This will make it easier to detect small syntax issues or other oversights in patch changes that can cause lines to not work at all that can be missed in review.

It adds a new yaml file for creating the action and configuring it to work on PRs, and a Python script that gets run by the Action job to do the actual validation. The script runs against every patch in the folder, which takes about ~4-5 seconds for the full Action to run.

### Coverage Checks
Currently it will detect the following issues:
- Patches that are missing fields, have an invalid field value,  or are otherwise malformed. (Including incorrect casing)
- Patch group headers straight brackets unclosed e.g. `[name`
- Non-lowercase `gametitle`, `comment`, `author`, `gsaspectratio`, `gsinterlacemode`, `description` keywords (if case is incorrect, these lines will get ignored)
- Broken comments at the end of a patch line (which break the line) e.g. `,00000001 /woops`
- Invalid comment lines/other garbage lines (of which many will be cleaned up with this PR).

### Patch Changes:
- Many cases of fixing comment syntax or removing single letter invalid lines.
- Changes numerous instances of  `1gametitle` to the proper `gametitle`
- Changes an instance of `gsaspectrpatio` to `gsaspectratio` so that this line will now work.
- Changes instance of "Author" to "author" so the line will not get ignored.
- Changes instance of patch with "Word" to "word" so the patch line will not get ignored.
- Fixes an instance if a ` in the middle of a patch line causing the line to get ignored.
- Fixes an instance of the `=` missing in `comment=`

### Notes;
- ~~Until #303 is resolved, it will detect 2 errors related to current patches using invalid `patch=3` syntax.~~ This patch was now fixed, and the Action is now passing due to finding 0 errors after the changes in this PR are applied.
<details>
<summary>No longer current image as these errors are now resolved, but shows how the action reports issues.</summary>

</details>

- Given the number of patch changes, it may be easier to review this with the 3 different commits as one is the CI changes, another is entirely comment changes, and the third being all the patch changes that impact behavior or otherwise fix lines that were getting ignored previously.


patches/SLKA-15042_BC0F422C.pnach
```diff
@@ -6,6 +6,6 @@ comment=Widescreen hack by Arapapa
 
 //Widescreen hack 16:9
 
-patch=1,EE,00C93688,Word,3F400000
+patch=1,EE,00C93688,word,3F400000
```
Daniel-McCarthy, line 9: Note: This line was previously being ignored entirely due to the uppercase W, so this change will have a behavior impact.

patches/SLPS-20476_715D9DD3.pnach
```diff
@@ -8,7 +8,7 @@ comment=Widescreen Hack by Arapapa
 
 //X-Fov
 //803f043c 7f4b023c
-patch=1,EE,001996a0`,word,3c043f40 //3c043f80
+patch=1,EE,001996a0,word,3c043f40 //3c043f80
```
Daniel-McCarthy, line 11: Note: This line was being skipped due to error with the syntax, so this will have a behavior impact.

pgert:
SCES-50244_71E38F05.pnach
enable "patch=1,EE,0014a04c,word,3c013f40"

SLPM-66211_7EC8A8A3.pnach
"gametitle=eter	..." ->  "gametitle=Peter ..."

Mrlinkwii (approved):
looks great

Daniel-McCarthy:
> Thanks for this :)
Of course. I'll be keeping an eye on it as PRs come in, if you see any issues or even have suggestions on checks to add, feel free to ping me. 👍

PR #312, opened 2024-02-10
Tomb Raider formating & improvments

pgert:


pgert:
Both Legend and Anniversary are unplayable with later builds;
 the physics are all wracked.
This has nothing to do with widescreen hacks,
 nor with Automatic Gamefixes (gsHWFixes).
And it's nothing I can do anything about.

PR #313, opened 2024-02-11
Ultimate Spiderman: widescreen disable, missing render fix and remove SLES-55571

PeterDelta:
Widescreen disabled for Ultimate Spiderman because the render fix is missing and it gives the impression that the game is unfinished or that pxsc2 does not render it well

Ultimate Spider-Man_SLES-53391_20240210182131

Ultimate Spider-Man_SLES-53391_20240211151648
Ultimate Spider-Man_SLES-53391_20240211151646
_______________________________________________________________________________________________________________________________________

Ghostbusters: SLES-55571_7030A01C.pnach removed because the region code does not exist
https://github.com/PCSX2/pcsx2_patches/issues/314


refractionpcsx2:
are we sure that doesn't exist? Redump seems to think so, they even dumped it http://redump.org/disc/11133/

PR #320, opened 2024-02-20
Fixes commented patch & incorrect game title

Daniel-McCarthy:
### Description:
- Fixes gametitle which had a typo, missing the P in Peter.
- ~~Fixes a commented patch which was the only patch in the pnach file, meaning the widescreen patch did nothing.~~ Removed this change as it is impossible for me to test currently. For anyone interested, [this line is here](https://github.com/PCSX2/pcsx2_patches/blob/955a9d824f8bf898618d582ba0136496102157f6/patches/SCES-50244_71E38F05.pnach#L10).

### Note:
Full credit to @pgert for reporting these in #304. I did not include those in those changes as these were both issues that would be undetected by the CI action so did not want to make it confusing.

### DIFF
--- a/patches/SLPM-66211_7EC8A8A3.pnach
+++ b/patches/SLPM-66211_7EC8A8A3.pnach
@@ -1,4 +1,4 @@
-gametitle=eter Jackson's King Kong - The Official Game of the Movie (J)(SLPM-66211)
+gametitle=Peter Jackson's King Kong - The Official Game of the Movie (J)(SLPM-66211)
 
 [Widescreen 16:9]
 gsaspectratio=16:9


Mrlinkwii (approved):

refractionpcsx2:
Do we know the patch you uncommented wasn't broken? has it been tested?

Daniel-McCarthy:
> Do we know the patch you uncommented wasn't broken? has it been tested?

Actually no, I don't know that for certain. The patch appears to have been commented out since it was added in the first place, so if no one has never noticed or complained about it not working, it probably doesn't hurt anything to simply delete, or move this to an Issue until it is able to be tested/verified.

refractionpcsx2:
I would like it either tested or removed, to be honest. I don't feel comfortable just turning it on when it was disabled to begin with. Maybe you can find someone with the game :D

Daniel-McCarthy:
I don't have this game and I'm not sure the likelihood of finding someone with this particular football game. I have gone ahead and removed it from this PR so at least the game title can be fixed.

Just because I think it's worth mentioning, that commented patch is the only patch in the file, and also the only patch line in the file. So the entire patch file does nothing and has been the case since the original commit.
I might make sense to simply remove the patch file entirely as it serves no purpose and hasn't. Just calling it out as it's unlikely to get examined again.

PR #324, opened 2024-02-21
Remove unnecessary No-Interlacing patches

PeterDelta:
As of this version: https://github.com/PCSX2/pcsx2/pull/10821 Patches can no longer be disabled per game. These types of patches that are only limited to eliminating the sceGsSetHalfOffset routine were a good fix in old versions but now they have become obsolete, do not improve anything and do worsen the textures and letters, so to keep the games with the least alteration I remove. following that I have verified are no longer necessary.

In Gran Turismo 3 we see that in automatic and the default interlacing modes of pcsx2 they do not act correctly so a patch would be needed.
Gran Turismo 3 automatic

Old patch that eliminates the routine sceGsSetHalfOffset 
Gran Turismo 3 routine interlacing

New patch enables progressive interlacing, keeps the sceGsSetHalfOffset routine unchanged, the image does not shake and the scrolling of the letters looks correct. The difference is minimal but it is there.
Gran Turismo 3 progressive

### DIFF
--- a/patches/SCES-50294_B590CE04.pnach
+++ b/patches/SCES-50294_B590CE04.pnach
@@ -1,14 +1,14 @@
 gametitle=Gran Turismo 3 - A-Spec (PAL-M) SCES-50294 B590CE04
 
 [No-Interlacing]
-author=junfurano
-description=Attempts to disable interlaced offset rendering.
 gsinterlacemode=1
-patch=1,EE,2025B4E8,extended,34020001
+author=PeterDelta
+description=Enable progressive interlacing.
+patch=1,EE,0028F81C,word,3C050000
 
 [60 FPS]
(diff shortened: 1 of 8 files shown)


Mrlinkwii:
patches can be disbled on a per game bases but you need to use teh patch menu

refractionpcsx2:
What? I just removed the option to disable the per game settings, control to enable and disable patches per game is still there.

refractionpcsx2:
Then the patches should be fixed. But I am aware you either enable them globally or only per game, there's no exception option. We've been discussing that.

The reason I said that was because you linked to my pr removing the option to disable per game settings, which is not the same thing.

refractionpcsx2:
Alright, cool thanks

PR #331, opened 2024-03-05
Update my FPS patches and widescreen

### LINKED ISSUE #356: [BUG]: FlatOut: Blurred textures when using aspect ratio fixes
When using widescreen patch builded in emulator or ultrawide patch i find in internet. Textures dont load or something very wrong happened with them. Also WS patch reduce trees drawing :( I check different API, different resolution with native, different setting and nothing but turning off mipmapping can resolve that. I
Original:
With build in emulator widescreen patch:

Also turning off texture filtering fix trees:  
FlatOut_SLUS-20901_20240518114047
Make them less accurate but crisper
And turning off mipmapping fix tires and further textures: 
FlatOut_SLUS-20901_20240518114053
 I would argue that that sacrifice is worth it.

Latest dev version of pcsx2, NTSC game, default setting with high blending like emulator purpose, i7 770, win 10.

https://youtu.be/jI658H3QUB0

### LINKED ISSUE #390: Crash Nitro Kart SLES-51511 480p patch
This is the patch published. Example in hell islands

[480p Mode]
gsinterlacemode=1
patch=1,EE,00446A74,word,24110000
patch=1,EE,00446A78,word,24120050
patch=1,EE,00446A84,word,24130001
patch=1,EE,0058BF7C,word,3C023EE8
Crash Nitro Kart_SLES-51511_20240713171243

This is image original at 50hz
Crash Nitro Kart_SLES-51511_50hz

Without the last line
Crash Nitro Kart_SLES-51511_sin linea

As it is, moving the image a little up would match the 50 Hz one, the only thing the hud would be a little bigger but I think it would be the least important.

### LINKED ISSUE #395: Need for Speed Underground (SLUS-20811) Widescreen Patch Possibly Not Working
Put a "possibly" there because honestly I'm not entirely sure myself, unlike I was with HP2.

But the WS patch does not actually seem to do much, if anything. If I switch to 4:3 and then back to 16:9, the game just stretches out, and none of the UI elements seem to adapt to the new format either.

Back in Hot Pursuit 2, that patch seemed to at least affect the menus. But here, I'm not seeing anything. I tried reloading the patches, checking and unchecking the patch in the Game Properties window while the game is running, and reloading the patch seems to make no difference at all. The 60fps patch does work correctly, but not the WS one.

How the game looks like on my end. The UI in particular is very stretched:

Need for Speed - Underground_SLUS-20811_20240720183800

PeterDelta:
I have already seen several users on the forum who are not clear about which 50/60 FPS patches to use in PAL versions. As we have added several patches to boot into NTSC or 480p video modes may cause confusion, I have simplified it and now it is clear, one patch to unlock FPS, either at 50 or 60, and another for video mode, either NTSC mode or 480p etc. I have updated some of them since they greatly increased the resources of the EE and now they perform better.

Other relevant changes aside:
Rogue Trooper: old patch is missing render fix, fixed

Rogue Trooper_SLES-52921_20240305110853

Rogue Trooper_SLES-52921_20240305110746

Area 51 SLES-53070 does not exist: There are only 3 PAL versions, SLED-53440, SLES-52570 and SLES-53075
http://redump.org/discs/quicksearch/area-51/sort/system/

Mercenaries 2: Old widescreen gives graphical glitches, fixed

Mercenaries 2 - World in Flames_SLES-55001_20240304161036
Mercenaries 2 - World in Flames_SLES-55001_20240304161431

Fixes #390 
Fixes #356 
Fixes #395


pgert:
"Unlock FPS" would i my opinon be better in "description" as "Unlocks FPS".
"[50 FPS]", "[60 FPS]", or "[50/60 FPS]" tells more about what the hack does (better for end-users).

"Might need EE Overclock to be stable" is more of a comment than a description, yes?

CRASHARKI:
I tested the Crash Bandicoot patches and I unfortunately found some issues: Crash Nitro Kart and Crash Tag Team Racing seem to be zoomed-in (maybe just vertically), I specially noticed it on gameplay (Race in CNK and Platforming in CTTR) rather than the menus. Crash Twinsanity has issues with the FMVs, which are played while the game is running on the background forcing them to run 5 fps faster and therefore skiping audio from time to time. I tested one version of Crash of the Titans and it ran fine, the screen is finally fixed.

CRASHARKI:
Hey I felt bad for being such a party pooper, so I researched a bit and found the correct value for the CNK patch, you just have to change the last line that fixes the height of the screen to this and you get a perfect result: patch=1,EE,0058BF7C,word,3C023F00

pgert:
If by "black bands" you mean blackbars, then I suggest you name it "blackbars" (which is 'standard'),
 and that you place the hacks seperately, as in [Remove Blackbars].

RedDevilus:
Not sure how Unlock FPS is less confusing than 50/60 FPS. If I see Unlock FPS I think of run as fast as possible like when you press F4 to uncap.

Mrlinkwii:
please leave it  to say 60 fps /50 fps :)

RedDevilus:
This is why you use the description section. You are stepping backwards creating a misleading term, if users don't get it even with revised and better description then it's not up to PCSX2 to make it easier for a minority but make it less obvious for majority, even PCSX2 forum calls it 50/60 FPS in the thread and other patches that are preincluded are also called that.

kamfretoz:
> Maybe it's better to say in description that it doubles internal game FPS like 25/30 to 50/60?

That, and also tell the user whether the patch needs a specific EE overclock level to be stable

patches/SLES-50335_7DB392C5.pnach
```diff
@@ -1,6 +1,14 @@
...
\ No newline at end of file
+description=Unlocks internal FPS. Might need EE Overclock at 180%.
+
+[480p Mode]
+gsinterlacemode=1
+author=PeterDelta
+description=Forces progressive scan mode 480p at startup.
+patch=1,EE,00367134,word,24110000
+patch=1,EE,00367138,word,24120050
+patch=1,EE,00367140,word,24130001
```
pgert, line 14: Correct me if I'm wrong, but these patches should not be included here, only in the secondary pnach.

patches/SLES-50382_6B149273.pnach
```diff
@@ -1,13 +1,13 @@
-gametitle=Silent Hill 2 (PAL-M6) (SLES-50382)
+gametitle=Silent Hill 2 (PAL-M6) (SLES-50382) 6B149273
 
 [Widescreen 16:9]
 gsaspectratio=16:9
 author=ElHecht
-comment=Widescreen Hack
+description=Widescreen Hack
 patch=1,EE,210E4C14,extended,43A8F5C3 // 43E147AE X-RES
 //patch=1,EE,210E4C18,extended,43E00000 // 43E00000 Y-RES
```
pgert, line 8: ElHecht made a new WS hack after this one:
patch=1,EE,001773C8,word,3c0243c0 // 3c024400 - X-FoV

The post seems to have been removed, though.

patches/SLES-53196_E3E8E893.pnach
```diff
@@ -1,6 +1,13 @@
 gametitle=Destroy All Humans (PAL-M) SLES-53196 E3E8E893
 
-[50 FPS]
+[Widescreen 16:9]
+gsaspectratio=16:9
 author=PeterDelta
-comment=Unlocked at 50 FPS. Might need enable EE Overclock to be stable.
+description=Enable native widescreen
+patch=1,EE,003F0404,word,3FE38E38 //3FAAAAAA
```
pgert, line 7: Sure it's not "3FAAAAAB" instead of "3FAAAAAA"?
And if so:
3FAAAAAB = 1,333333
(1+(1/3))/0,75 = 1,77777777777777777778 = 3FE38E39
"3FE38E38" -> "3FE38E39".

patches/SLES-55639_F4299325.pnach
```diff
@@ -0,0 +1,21 @@
+gametitle=Ben 10 - Ultimate Alien - Cosmic Destruction (PAL-M) SLES-55639 F4299325
+
+[Widescreen 16:9]
+gsaspectratio=16:9
+author=PeterDelta
+description=Renders the game in 16:9 aspect ratio
+patch=1,EE,00765154,word,3FE38E38 //3FAAAAAB
```
pgert, line 7: "3FE38E38" -> "3FE38E39".

patches/SLES-55470_4F82849C.pnach
```diff
@@ -14,7 +14,15 @@ patch=1,EE,001A6548,word,3C013C2E //3C013C0E
 patch=1,EE,00225ea8,word,3c014000 //3c014299
 patch=1,EE,00225eac,word,00000000 //3421999a
```
pgert, line 15: If you're up for it make this a hack of its own, as in "[Remove Blackbars]".

pgert:
Nothing more to add.
This PR looks good to me.

CRASHARKI:
You might as well re-add the fixed Crash Nitro Kart 480p mode patch, right?

CRASHARKI:
> It is not fixed, the only thing left to correct is what you mentioned about the vertical axis of the image because its resolution increases to 576 and therefore the viewing area increases. I've been looking at it and so far I haven't found anything... It can be played well, I have enjoyed several circuits without problems but since the image is not the same as in 50 Hz I am not sure about including it

Wasn't it fixed by just getting rid of the patch=1,EE,0058BF7C line in the patch? It looks good to me, no stretch issues. Can you specify in which case/level are you experiencing issues?

patches/SLES-52646_CE64E328.pnach
```diff
@@ -3,13 +3,13 @@ gametitle=Tom Clancy's Ghost Recon 2 (PAL-M) SLES-52646 CE64E328
 [Widescreen 16:9]
 gsaspectratio=16:9
 author=PeterDelta
```
pgert, line 5: author=arapapa
https://forums.pcsx2.net/Thread-PCSX2-Widescreen-Game-Patches?pid=610600#pid610600

pgert:
Does "gsinterlacemode=1" make any difference in a "[480p Mode]" hack?
Isn't it of consequence only in a "[No-Interlacing]" hack?

Mrlinkwii:
like as mentioned please dont use the word unlock

patches/SLES-52379_36D7FE6C.pnach
```diff
@@ -0,0 +1,15 @@
...
+gsaspectratio=16:9
+author=PeterDelta
+description=Renders the game in 16:9 aspect ratio
+
+[50/60 FPS]
+author=PeterDelta
+description=Might need EE Overclock at 130%.
+
+[480p Mode]
+author=PeterDelta
```
pgert, line 14: Isn't "gsinterlacemode=1" required here?

patches/SLES-52382_4CF48A70.pnach
```diff
@@ -5,5 +5,13 @@ gametitle=Shrek 2 (Spain) (SLES_523.82)
...
\ No newline at end of file
+author=PeterDelta
+description=Renders the game in 16:9 aspect ratio
+
+[50/60 FPS]
+author=PeterDelta
+description=Might need EE Overclock at 130%.
+
+[480p Mode]
+author=PeterDelta
```
pgert, line 16: Same thing here.

PR #338, opened 2024-03-29
WS: Disable SMT DS 2 WS patch

JordanTheToaster:
Breaks sprite overlap in some scenes.

image


pgert:
Suggestion:
Instead of disabling "black borders's fix", make it a hack of its own, as in [Remove Blackbars].

PR #353, opened 2024-05-15
Update SLUS-20675_B0859096.pnach

Joshua-1248:
Adds deinterlacing patch to existing pnach for Baldur's Gate - Dark Alliance II.


Joshua-1248:
> These types of patches have been eliminated because they make them look worse, they are no longer necessary. If you do a comparison it is clear.
> 
> Baldurs gate 2
> 
> With (left) and without (right) deinterlacing patch

I see what you mean there.

I'll see if I can figure out a way to disable interlacing without having the half-offset visual anomalies.

Closing.

PR #366, opened 2024-05-28
Adding SLES-51194 Pnach 

Snake356:
There is no Pnach file for SLES-51194 yet and has been added now. The Pnach contains exactly the same as for SLES-51195, as it is identical to the CRC, it is just a different language of the game.

### DIFF
--- a/patches/SLES-51194_C5473413
+++ b/patches/SLES-51194_C5473413
@@ -4,6 +4,7 @@ gametitle=Harry Potter und die Kammer des Schreckens (G)[SLES-51194]
 gsaspectratio=16:9
 author=Arapapa
 comment=Widescreen Hack
+
 //Widescreen hack 16:9
 //Zoom
 patch=1,EE,004ca798,word,3c013f24 //3c013f49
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
The fact half of it is commented out because it's broken, but was there because it needed it bothers me.  If this patch breaks some of the visuals, I don't really want to supply it with the emulator.

Snake356:
> The fact half of it is commented out because it's broken, but was there because it needed it bothers me. If this patch breaks some of the visuals, I don't really want to supply it with the emulator.

I haven't had any problems with the game at all, so I've played through several times with the patch and haven't noticed any graphical errors. This patch is also offered here for the Spanish version, so I can't understand why the German version can't be offered for it. Because then you would also have to remove the patch for the Spanish version, because it contains the same code.

refractionpcsx2:
As long as you're sure the UI stuff is all in the right place, then that's fine.

Snake356:
> As long as you're sure the UI stuff is all in the right place, then that's fine.

Everything looks correct, see the 2 pictures I have added.
 
Screenshot 2024-05-28 232131

Screenshot 2024-05-28 232312

refractionpcsx2:
the UI all looks stretched, not 16:9

Snake356:
> the UI all looks stretched, not 16:9

Yes you are right was stretched, have now examined the code again exactly and the commented code times tried everything looks correct with the UI now see pictures. I have also changed the code again now.

Screenshot 2024-05-28 233630

Screenshot 2024-05-28 233658

refractionpcsx2:
Did you just re-enable the code? cos if it was disabled for reasons then those reasons are probably still a thing. 

I have to be honest, my confidence in the patch (and your ability to modify it) is slightly worrisome right now. I'm not sure I want to merge this, at least in its current state without extensive testing.

Snake356:
> Did you just re-enable the code? cos if it was disabled for reasons then those reasons are probably still a thing.
> 
> I have to be honest, my confidence in the patch (and your ability to modify it) is slightly worrisome right now. I'm not sure I want to merge this, at least in its current state without extensive testing.

Yes, I just excluded it, it just says “// Commented out as it broke loading screen font and other stuff.” The only thing I noticed was that nothing is displayed at the beginning when you load a savegame. I had already used this patch several times and had also played like this and so far. I don't know if this problem can be fixed. 

Screenshot 2024-05-28 235409

refractionpcsx2:
I'm sorry, you'll have to get the author to fix it, I can't accept patches which break the graphics in any way.

Snake356:
> I'm sorry, you'll have to get the author to fix it, I can't accept patches which break the graphics in any way.

All right, no problem.

PR #367, opened 2024-06-01
Allow the Ratchet and Clank Widescreen Patch

sandboxgamedev123:
### Description

I am proposing to allow the widescreen patch for Ratchet and Clank.

The widescreen patch for Ratchet and Clank is unavailable to use.

The file is "SCUS-97199_CE4933D0.pnach".

The following is the comment in the patch file:

"Widescreen hack by PsxFan107. Currently it appears to break rendering of textures in some areas."

I edited the patch to allow it to work in the game. I played through sections of the game and I did not notice any notable texture issues as described in the comment.

URL: https://github.com/sandboxgamedev123/PCSX2-Ratchet-and-Clank-Widescreen

The URL refers to a repository I created that provides the edited patch file as well as additional information. I copied the code to the existing file in the "pcsx2_patches" repository.

### Reason

The widescreen patch allows the game to operate with a 16:9 aspect ratio in-game.

### Checklist

The test systems:

 - [X] AMD Ryzen 5 5600H, NVIDIA GeForce RTX 3060 6 GB, 16 GB RAM, 1.5 TB NVMe SSD
 - [X] Intel Core i5-7300HQ, NVIDIA GeForce GTX 1050 Ti 4 GB, 16 GB RAM, 512 GB SATA SSD

### DIFF
--- a/patches/SCUS-97199_CE4933D0.pnach
+++ b/patches/SCUS-97199_CE4933D0.pnach
@@ -1,101 +1,99 @@
-//gametitle=Ratchet & Clank (NTSC-U)
+gametitle=Ratchet & Clank (NTSC-U)
 
-//[Widescreen 16:9]
-//gsaspectratio=16:9
-//comment=Widescreen hack by PsxFan107. Currently it appears to break rendering of textures in some areas.
+[Widescreen 16:9]
+gsaspectratio=16:9
+comment=Widescreen hack by PsxFan107. Currently it appears to break rendering of textures in some areas.
 
 // DWORD Code patching routine
-//patch=1,EE,200C0000,extended,3C1B000C
-//patch=1,EE,200C0004,extended,DF710080
-//patch=1,EE,200C0008,extended,14510002
-//patch=1,EE,200C0010,extended,DF620088
-//patch=1,EE,200C0014,extended,DF710090
-//patch=1,EE,200C0018,extended,14510003
-//patch=1,EE,200C0020,extended,8F7200A0
-//patch=1,EE,200C0024,extended,AC72FFF0
-//patch=1,EE,200C0028,extended,DF710098
-//patch=1,EE,200C002C,extended,14510002
-//patch=1,EE,200C0034,extended,AC72FFF0
-//patch=1,EE,200C0038,extended,0804B657
-//patch=1,EE,200C003C,extended,FC62FFF8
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Can you show this area with your patch? I'm not quite sure which one it is https://github.com/PCSX2/pcsx2_patches/pull/74

refractionpcsx2:
Come on mate, we're not fucking stupid. You just re-enabled the old patch, I can compare files, you know. Stop wasting our time.

sandboxgamedev123:
I did not say the patch code was different. I stated that the patch was edited to allow it to work in-game. I was in the area from the link you provided and the area did not look like the image. There was occasional flickering on the foliage but I do not know if it is a LOD issue or texture issue. I think it is much more minor than what was shown in the image. And as I mentioned before, I tested sections of the game on two systems and I think it generally looked as it should. I was noticing texture detail transitions near that area. In this case I am discussing a 2002 game operating in HD so flaws in the rendering might be more noticeable. I think the issues were very minor.

I will provide screenshots of the area.

I used the Vulkan renderer. I do not know if other renderers produce issues. Regardless, I think Vulkan is important for operating the emulator on modest hardware so the other renderers could be considered less relevant.

refractionpcsx2:
There's no point checking it now, it's the same patch, it's still going to be broken. I'm not accepting it, it is broken.

sandboxgamedev123:
Why do you believe is it broken? It does not look broken to me. How many people reported issues with it?

refractionpcsx2:
We had several people report issues with it, to the point we looked in to it and discovered it was the patch.  We haven't had a single report since.

sandboxgamedev123:
The following are images from a system with an AMD Ryzen 5 5600H and a NVIDIA GeForce RTX 3060 using the Vulkan renderer.



















As I think you can see, I think the game generally looks as it should. If you would like I can conduct the test again with the AMD Radeon Vega 7 integrated GPU, the Intel i5-7300HQ CPU, and the Intel HD Graphics 630 integrated GPU.

sandboxgamedev123:
The following are two more images in a new comment because there seemed to be an issue putting more than 7 images in one comment.











I would be surprised that the patch generally works well on my range of hardware but yet you see reports from certain people regarding the patch. The commonality with my hardware is full Vulkan 1.3 support. I am wondering if the issue is related to the renderer being used or GPU renderer compatibility.

PR #370, opened 2024-06-03
Urban Reign (NTSC) SLUS-212.09 - Deinterlacing & De-Blur Patches

Joshua-1248:


Joshua-1248 (commented):
Urban Reign (NTSC) SLUS-212.09 - Deinterlacing & De-Blur Patch

Joshua-1248:
> Please, could you share a screenshot of each game with the patch enabled and another with the default pcsx2 options?, so we can see if there is improvement or if it gets worse. Basically you are removing the sceGsSetHalfOffset routine and in the latest versions of pcsx2 in most games the image does not shake and runs fine, therefore removing those values ​​makes the image look worse. #353

These patches are different from NOP'ing the sceGsSetHalfOffset instruction routine as they are placed next to the interlacing address in memory which cycles between 0 and 1, & back to 0 & then 1 again.

There happened to be two settings there that I could just switch the value of to zero to disable not only the interlacing, but the screen blur.

Urban Reign_SLUS-21209_20240603152812
Urban Reign_SLUS-21209_20240603152817

pgert:
Upload the pnachs directly to the 'patches' folder of the repo.
Don't have a 'cheats_ni' folder in which you put them.

If a pnach already exists for the disc, then download the repo as a zip-folder,
 extract and find the pnach, add the hack to that pnach, and upload that pnach.

Don't have "//Deinterlacing Patch" or "//Deinterlacing Code" entries.
Make them hacks that can be turned on/off,
 as in "[Deinterlacing Patch]" & "[Deinterlacing Code]".

BetterWarrior:
Thank you Joshua, for this patch it's really a game changer i was tired of trying to fix the game i tried x8 resolution i tried all deinterlacing methods for the game, i even tried x8 upscale of the game but it was blurry and shaken but with your patch the game look really incredible.

BetterWarrior (approved):

BetterWarrior (approved):

Mrlinkwii:
whats the status of the pr?

PR #372, opened 2024-06-06
Add many 60 FPS patches

icup321:
Adds many patches from various sources, mainly asasega's 60 FPS patches and NTSC versions of PeterDelta's patches.


Mrlinkwii:
are any of these tested ?

icup321:
> are any of these tested ?

The vast majority of these have been tested to work. Most of these games share the same engines so I've at least had the codes tested on each game engine. I also had to modify some of them to get them to work properly as well.

Mrlinkwii:
@icup321  can you rebase this pr for me

icup321:
Should be rebased now, can't squash all commits but shouldn't be a big deal

PR #373, opened 2024-06-06
Allow the Ratchet and Clank Widescreen Patch

sandboxgamedev123:
It is possible that the previous pull request was closed prematurely.

This is based on the following issue:

https://github.com/PCSX2/pcsx2/issues/11321

This is also based on the following pull request:

https://github.com/PCSX2/pcsx2_patches/pull/367

### Checklist

Operating systems tested:

 - [X] Windows 10 1903 64-bit
 - [X] Linux Mint 21.2 Cinnamon 64-bit

Test systems:

 - [X] AMD Ryzen 5 5600H, NVIDIA GeForce RTX 3060 6 GB, 16 GB RAM, 1.5 TB NVMe SSD
 - [X] Intel Core i5-7300HQ, NVIDIA GeForce GTX 1050 Ti 4 GB, 16 GB RAM, 512 GB SATA SSD

The following is the most recent comment from the previous pull request:

The following are images from PCSX2 v1.7.5858 on a system with an AMD Ryzen 5 5600H and a NVIDIA GeForce RTX 3060 using the Vulkan renderer.

As I think you can see, I think the game generally looks as it should. If you would like I can conduct the test again with the AMD Radeon Vega 7 integrated GPU, or the Intel Core i5-7300HQ CPU with the Intel HD Graphics 630 integrated GPU or the NVIDIA GeForce GTX 1050 Ti.

Since the patch generally works well on my range of hardware but yet you saw reports from several people regarding the patch, I am wondering if it is not specifically an issue with the patch. The GPU API commonality with my hardware is Vulkan 1.3 support. I am wondering if the issue is related to the renderer being used or GPU API compatibility.

I am also wondering if the issue was already fixed from updates to the emulator. What was the newest report you received? I have been using PCSX2 again since v1.7.5290 and I did not notice the issue in v1.7.4721 either.

### DIFF
--- a/patches/SCUS-97199_CE4933D0.pnach
+++ b/patches/SCUS-97199_CE4933D0.pnach
@@ -1,101 +1,99 @@
-//gametitle=Ratchet & Clank (NTSC-U)
+gametitle=Ratchet & Clank (NTSC-U)
 
-//[Widescreen 16:9]
-//gsaspectratio=16:9
-//comment=Widescreen hack by PsxFan107. Currently it appears to break rendering of textures in some areas.
+[Widescreen 16:9]
+gsaspectratio=16:9
+comment=Widescreen hack by PsxFan107. Currently it appears to break rendering of textures in some areas.
 
 // DWORD Code patching routine
-//patch=1,EE,200C0000,extended,3C1B000C
-//patch=1,EE,200C0004,extended,DF710080
-//patch=1,EE,200C0008,extended,14510002
-//patch=1,EE,200C0010,extended,DF620088
-//patch=1,EE,200C0014,extended,DF710090
-//patch=1,EE,200C0018,extended,14510003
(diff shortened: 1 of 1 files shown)


Mrlinkwii:
you didnt fix anything , it literally the same patch

sandboxgamedev123:
I not say the patch code was different. The title says "Allow the Ratchet and Clank Widescreen Patch".

refractionpcsx2:
> I did not say the patch code was different. The title says "Allow the Ratchet and Clank Widescreen Patch".

image

That sounds like you modified it to work, since it was disabled for being broken.

Anyways, we're looking in to it, we have reproduced the error again, but it may not be directly related to the patch, but the widescreen patch does make it worse.

sandboxgamedev123:
What are the specifications of the system where the issue is reproducible?

PR #382, opened 2024-06-28
Update SLES-51058_54854C71.pnach for proper PAL Progressive scan

### LINKED ISSUE #380: Maken Shao PAL (SLES-51058_54854C71) no interlacing code forces NTSC 60 FPS profile and immediately desyncs all FMVs
The Maken Shao patch provides a no-interlacing code but pushes a PAL Interlaced format into a NTSC Progressive format, making it into a 60 FPS game too. It desyncs every FMV by playing them too fast and considering every cutscene is pre-rendered, this pretty much makes the story impossible to follow.

The description of the no-interlacing code doesn't provide any mention of a 60 FPS framerate mode coming with the no-interlacing option. I think this will make it a worse experience for players to leave it as is

```
[No-Interlacing]
gsinterlacemode=1
comment=Progressive scan

//Override SetGsCrt syscall arguments.
patch=1,EE,202000c0,extended,24040000
patch=1,EE,202000c4,extended,24050002
patch=1,EE,202000c8,extended,24060001
```

Removing the second line `patch=1,EE,202000c4,extended,24050002` brings it back to a PAL Progressive format and removes the 60 FPS NTSC format stuck in there. As long as FMVs remain desynced without an included fix, I think this line should be removed from the patch and have a more accurate comment description to users. Sorry I'm not familiar enough to make the fixes myself

Wiinansa:
PAL Progressive scan instead of NTSC Progressive Scan forcing 60 FPS on the user and having the consequence of de-syncing all Maken Shao cutscenes to end faster than they should.

It should resolve #380

### DIFF
--- a/patches/SLES-51058_54854C71.pnach
+++ b/patches/SLES-51058_54854C71.pnach
@@ -12,8 +12,6 @@ comment=Progressive scan
 
 //Override SetGsCrt syscall arguments.
 patch=1,EE,202000c0,extended,24040000
-patch=1,EE,202000c4,extended,24050002
+// The next line activates 60 NTSC Progressive format but brings a slew of issues like immediate FMV desync and the game being out of frame unless using the overscan option
+// patch=1,EE,202000c4,extended,24050002 
 patch=1,EE,202000c8,extended,24060001
-
-
-


Mrlinkwii (approved):
assumed tested

PR #385, opened 2024-06-30
Add patches for RA games

MrWrightyLC:
This PR adds support for patches for some games that are supported on the RetroAchievements site (sometimes some RetroAchievements sets use some games that are modified by xdelta patches, which causes a CRC change)
So using the patches from this same repository of the regular versions to support these versiones with a simple CRC change

- [Raw Danger! | Zettai Zetsumei Toshi 2: Itetsuita Kiokutachi](https://retroachievements.org/game/5616)
adds support for WS
RA Raw Danger

- [Shin Megami Tensei: Persona 3 FES](https://retroachievements.org/game/2657)
adds support for WS, Mode 480p and additionally [Controllable Party Members] (Only for 94A82AAA and 97102837)
RA Persona 3 Fes

I have completed both games to make sure they work.
Persona 3 Fes with a 143+ hour game (using Controllable Characters UNDUB - Original BGM) and I made sure to play the other versions as well for at least 40 minutes.
Raw danger was also played in a game of more than 9 hours (using NTSC Mode + Performance) and I also played the other version for 1 hour.


F0bes:
Could you make the patches use the `author` field instead of putting the name in the comment?

PR #398, opened 2024-07-30
Update SCUS-97328_77E61C8A.pnach

regiregi22:
Patch for Color depth will remain 32-bit while using 480p.

### DIFF
--- a/patches/SCUS-97328_77E61C8A.pnach
+++ b/patches/SCUS-97328_77E61C8A.pnach
@@ -116,3 +116,16 @@ comment=The game will always start in 480p instead of 480i.
 author=Silent
 patch=1,EE,204364A8,extended,AE0516B0
 patch=1,EE,10436598,extended,10E8
+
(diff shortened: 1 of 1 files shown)


regiregi22:
There is some more info here about this patch:
https://www.youtube.com/post/UgkxOeBlZUW4xh_p5e0_TskC8xsFfDy-X32W

regiregi22:
This is the source from the creator:
https://github.com/AeroWidescreen/PCSX2-Cheats/blob/main/Gran%20Turismo%204/SCUS-97328/32-Bit%20Color%20Depth/SCUS-97328_77E61C8A_colordepth.pnach

pgert:
#348 : "causes the blue rectangle of death on the left hand side of the screen".

regiregi22:
> #348 : "causes the blue rectangle of death on the left hand side of the screen".

Thank you for that, I will then close the PR!

PR #403, opened 2024-08-11
Blur post processing removal patches for Prince Of Persia games

F0bes:


pgert:
"[Remove Blur/Bloom]" would be clearer (without whitespaces).

F0bes:
> "[Remove Blur/Bloom]" would be clearer (without whitespaces).
Sounds good to me.

PR #404, opened 2024-08-15
No Interlacing Patch for Extermination [SCUS-97112]

Snake356:
Added No Interlacing Patch from asasega for Extermination. The game shakes a lot without no interlacing, that's why I added the patch from asasega, because it hasn't been added to the patches yet. Also works without any problems.


Snake356:
Oh I see, yes I didn't know that something like that was no longer needed then of course I will close the whole thing.

NineKain:
In my case, when moving you get crazy shimmering with it on "Automatic", disabling it and using my deinterlacing patch fixes it.

PR #411, opened 2024-09-01
Men in Black II Widescreen disabled 

PeterDelta:
The widescreen patch breaks the game's textures continuously when moving the camera

Men in Black II - Alien Escape_SLES-50789_20240902001212
widescreen on

Men in Black II - Alien Escape_SLES-50789_20240902001159
4:3 original


F0bes (approved):
Thanks

PR #413, opened 2024-09-05
WS: Disable Battlegear 2 WS patch

JordanTheToaster:
Break car that is bad.

image


F0bes (approved):
You'd think they'd notice this when they made the patch?

PR #414, opened 2024-09-05
Corrected Tekken Tag Tournament (SLUS-20001) character ratio

crashGG:
The ratio of Tekken TT should be 640:448 (original pixel ratio), but it is usually stretched to 4:3 in pcsx2, causing the y-axis to be stretched by 7.14% and the character to look a bit thin and long visually.This issue also carries over to using the 16:9 patch.

Now adjust a value on the 16:9 patch to get the character back to the correct ratio.

master:

pr:


crashGG:
The same for Tekken Tag Tournament (Japan) (v3.00)  ,and with a No-Interlacing patch added

patches/SLPS-20015_06979F19.pnach
```diff
@@ -3,8 +3,11 @@ gametitle=Tekken Tag Tournament [NTSC-J] (SLPS-20015)
 [Widescreen 16:9]
 gsaspectratio=16:9
 author=nemesis2000
-patch=1,EE,0034b004,word,3c013f40
+patch=1,EE,0034b004,word,3c013f4E
```
F0bes, line 6: Nitpick, can you make the E lowercase.
Same with the other patch.

Mrlinkwii:
@crashGG  can this be fixed if you have time :)

Mrlinkwii (approved):

F0bes (approved):

PR #425, opened 2024-10-04
Obscure: The Aftermath - Disable 60FPS patch

MrWrightyLC:
Reason: This patch breaks and prevents progress
As discussed on the RA site
Screenshot_2024-10-04-14-24-02-365_org mozilla firefox-edit


pgert:
Suggestion: Make the 480p hack a separate hack while at it.

Mrlinkwii:
theirs 0 reason to remove it , just because achievements hate it

PR #445, opened 2024-11-13
Correct and tested Power Rangers - Dino Thunder patches

Gabominated:
By mistake in my most recent merged pull I have used raw Power Rangers - Dino Thunder patches but here I bring the correct and tested ones

Testing patch
[https://youtu.be/PC8ZSiikusE](url)


Mrlinkwii (approved):
assumed tested

F0bes (approved):

PR #448, opened 2024-11-20
Fatal Frame 2 Undub - Patches update

MrWrightyLC:
- The undub patch was updated a few months ago and with that update the CRC changed again, so this PR adds support for this new Undub v1.0.2 (1C6C1B71) ~~basically using the same patches as Undub v1.0.1 (1C6C1DB6).~~
- So the support for the old v1.0.1 is removed, in order not to fill this repository with unnecessary CRC's, leaving only the support for the most recent version of this romhack.
- as well as this PR serves as a follow up of #424
since that PR forgot to disable the broken 60fps patch for the Undub patch

~~i would also like to have the opinion of a moderator~~

~~1. I should remove support for the old undub patch v1.0.1 (1C6C1DB6) from Nov 1, 2022 and only support for the latest undub patch v1.0.2 (1C6C1B71) from May 17, 2024?~~

~~2. Or leave it as it is for possible players that are still using the old undub v1.0.1?~~


Mrlinkwii (approved):

F0bes:
I should have commented out the patch instead of removing it. Can you just comment it out instead?

PR #450, opened 2024-12-01
Add patches and fixes

PeterDelta:
Midnight Club - Street Racing: Fixed car selection
Tekken 5 USA: Added 480p patch, enables native 480p mode from start
Kingdom Hearts II USA and JAP Final Mix+ version: 60 FPS patch was missing in these versions, now they are complete


Mrlinkwii (approved):
assumed tested

PR #451, opened 2024-12-03
Added 50 FPS Patch for the German Harry Potter and the Chamber of Secrets Version

Snake356:
Has been tested and works without problems!


Mrlinkwii (approved):
assumed tested

PR #468, opened 2024-12-30
Tenchu: Fatal Shadows - Add Undub patch

MrWrightyLC:
**Adds undub patch that forces Japanese voices at all times**

this game has a built-in option to activate the japanese voices, but this option always needs to be enabled manually each time the game is launched as it is disabled when the game is closed.
But with this patch it will always force the use of the japanese voices automatically.


Mrlinkwii:
has this been tested ?

PR #473, opened 2025-01-01
SSX 3 Aspect Ratio Patches

Virjoinga:
Sorry for deleting the recent pr, issues on my end :/

This only works for the NTSC-U version.

### DIFF
--- a/patches/SLUS-20772_08FFF00D.pnach
+++ b/patches/SLUS-20772_08FFF00D.pnach
@@ -56,11 +56,6 @@ description=Changes aspect ratio.
 patch=1,EE,20622600,extended,3F0EE23C //Horizontal Aspect Ratio
 patch=1,EE,20622604,extended,3F800000 //Vertical Aspect Ratio
 
-patch=1,EE,20622604,extended,3f800000 // 3f800000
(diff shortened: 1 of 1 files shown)


patches/SLUS-20772_08FFF00D.pnach
```diff
@@ -0,0 +1,69 @@
...
+author=Virjoinga
+description=Changes aspect ratio.
+patch=1,EE,20622600,extended,3F0EE23C //Horizontal Aspect Ratio
+patch=1,EE,20622604,extended,3F800000 //Vertical Aspect Ratio
+
+patch=1,EE,20622604,extended,3f800000 // 3f800000
+patch=1,EE,e0023f40,extended,006225FE // fix for vert- cutscenes
+patch=1,EE,20622600,extended,3f100000 //
+patch=1,EE,20622604,extended,3f400000 //
+
```
pgert, line 63: This "Cutscenes fix" is for 16:9.
Why have you included it here, and only here?

Virjoinga: Oh shoot, that was left in accidentally. I'll remove it.

PR #485, opened 2025-01-13
Red Ninja: End of Honor - Reduce Camera Sensitivity

MrWrightyLC:
This little patch slows down the insane sensitivity used in this game


Mrlinkwii:
i assume this has been tested

PR #486, opened 2025-01-15
GTA San Andreas Widescreen improvement and Silentpatch Lite for 399A49CA and 2C6BE434

MackdannyTwitch:
The current widescreen patch, while functional, has the HUD and the radar unaffected. I include pnach files for 1.03 and 3.00 that make the widescreen appear more like ThirteenAG's widescreen mod on PC, thus fixing the HUD and radar. It also has the added benefit of not glitching the emergency vehicles. I also include a Silentpatch lite that adds cosmetic and QOL enhancements, similar to the PC version, while leaving game mechanics otherwise untouched, so it wouldn't trigger RetroAchievements' anticheat system.

1.03
Grand Theft Auto - San Andreas_SLUS-20946_20250108112003

3.00
Grand Theft Auto - San Andreas_SLUS-20946_20250108130059

(For best results, turn the ingame widescreen option on, but that does not affect the radar/HUD whether on or off)

### DIFF
--- a/patches/SLUS-20946_2C6BE434.pnach
+++ b/patches/SLUS-20946_2C6BE434.pnach
@@ -2,24 +2,148 @@ gametitle=Grand Theft Auto: San Andreas (SLUS-20946GH) 2C6BE434 / Ver 3.00
 
 [Widescreen 16:9]
 gsaspectratio=16:9
-author=nemesis2000
-description=Widescreen fix
-patch=1,EE,001130BC,word,3C013F9D
-patch=1,EE,001130C0,word,44810000
-patch=1,EE,001130C4,word,46006302
-patch=1,EE,001130C8,word,03E00008
-patch=1,EE,001130CC,word,E78C9A90
+author=PeterDelta, DanielSantos, ThirteenAG, kesterstudios
+description=GTA San Andreas Modern Widescreen Fix for 3.00/GH 
 
-patch=1,EE,0021DFE4,word,0C044C2F
-patch=1,EE,00242DB4,word,0C044C32
+patch=1,EE,207004EC,extended,01000101 //Widescreen fix by PeterDelta
+patch=1,EE,0021E040,extended,3C0242CC
+patch=1,EE,0020A4F8,extended,3C0242AC
+patch=1,EE,002081DC,extended,3C0342AC
+patch=1,EE,002ECAF0,extended,3C044040
+patch=1,EE,002ECB40,extended,3C0440D0
+patch=1,EE,0033DB54,extended,3C013F80 //zoom in scenes
+patch=1,EE,E0010001,extended,0066C174
+patch=1,EE,0033DB54,extended,3C013F40
+patch=1,EE,2021EE14,extended,3C023FC0
+//-------------------------------------------------HUD-------------------------------------------------//
+patch=0,EE,20664384,extended,3ecccccd //Wanted Width
+patch=0,EE,20664390,extended,3ef5c28f //Wanted Shadow Width
+patch=0,EE,2026E09C,extended,3c084280 //Radar Width
+patch=0,EE,2026ED60,extended,3c024280 //Radar Mask Width
+patch=0,EE,2026EE90,extended,3c034280 //Radar Mask Width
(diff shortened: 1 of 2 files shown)


MackdannyTwitch:
> Hi mate, I've noticed a bug that is inherited from the nemesis code that this patch incorporates. When you go too fast, errors still appear on the sides, although they are less noticeable, they still persist. Regardless of the camera we choose.
> 
> Grand Theft Auto - San Andreas_SLUS-20946_20250118173306
> 
> Grand Theft Auto - San Andreas_SLUS-20946_20250118173029

Two questions, is this on both of the versions I've supplied patches to? And if so, since i know you had yours patched for that issue, i was wondering... this one and yours had the emergency vehicles patched, but while yours doesnt have the stuff on the sides, this one has a perfect HUD/RADAR... what if we could incorporate all of it? 

Presently, the 3 main games have a bad HUD/RADAR, but games like VCS amd LCS actually look great with the patches they have supplied.

CookiePLMonster:
Is this ready for review?

MackdannyTwitch:
I'm at a crossroads here.... on the one hand, everything i included, i fully stand by, and think it should be added to the game. On the other, I get what PeterDelta is saying and that the quirks of the old nemesis2000 work is in there. I do think he was onto something when he corrected it, however the other stuff that was corrected in my submission shouldn't be discarded either. I just.... dont know how to combine the best of both worlds. Any ideas?

CookiePLMonster:
@PeterDelta Would you be able to help out here in merging the fixes?

MackdannyTwitch:
Tested both. Nice! Wasnt the black outline on the font part of the SilentPatch thing?

CookiePLMonster:
> Tested both. Nice! Wasnt the black outline on the font part of the SilentPatch thing?

It's technically part of the new widescreen fix but that's because ThirteenAG, who made the PC fix, is obsessed with those. I would personally advise against including it as either part of SP Lite or the widescreen fix. I would be totally fine having it as a separate patch, though, since some people might find outlined fonts easier to read.

MackdannyTwitch:
I've isolated it. I'll follow your advice on the matter. I could have sworn it was under SP Lite, but i had seen it incorrectly. In any case, it shall be it's own thing.

CookiePLMonster:
One more comment - I'm pretty sure that the "Menu" changes belong to the widescreen fix, not SP Lite.

MackdannyTwitch:
I think Daniel Santos  or Kesterstudios had it separated, but if you think it belongs under the widescreen section, i can do that.

MackdannyTwitch:
Perspnally i do agree it does look better as part of the widescreen fix

MackdannyTwitch:
@PeterDelta is the Emergency vehicles fix still on here?

MackdannyTwitch:
@CookiePLMonster ready

CookiePLMonster (approved):
A few notes:

1. Tested 1.03, looks all good.
2. There's considerably less patches in 3.00, is this intentional or is the port partial?

MackdannyTwitch:
I noticed that too. It appears kesterstudios didn't port every single thing, but the bare essentials are definitely in there. The black outlines, for example, is one thing that didn't appear on 3.00. 

This is my source https://www.psx-place.com/threads/gta-sa-v1-03-ntsc-modern-widescreen-fix-project.35603/#post-380145

MackdannyTwitch:
Thanks! And thank you both for the help!

Mrlinkwii (approved):

ElTioRata:
Camera FOV on interiors is... a little too wide?
imagen

MackdannyTwitch:
If I had to hazard a guess, the gane insists on not letting you see outside the interiors, which is great because we know GTASA's interiors are actually at a very high point in the sky box, so we dont wanna see the sky. That being said, my guess is that the camera doesnt really follow CJ too much if the interior you're in is narrow. It catches up to CJ in the kitchen, as the kitchen is more spacious. Idk if this is possible to fix, and if it is, it's beyond my ability, but its also not too annoying, imo. I did compare it to the PC widescreen mod and in that one, the camera instead sticks too close to CJ when in narrow spaces (like some interiors). Good observation, though.

PR #491, opened 2025-01-29
WS/Misc: Various fixes

JordanTheToaster:
Removes DMC 2 WS due to causing SPS and removes both patches for Rule of Roses due to them breaking the airship map.

DMC 2 Issue:
image

Rule of Roses Issue:
image


patches/SLUS-21448_F3FD313E.pnach
```diff
@@ -19,13 +19,13 @@ patch=1,EE,0013bd7c,extended,3c013f10
...
 
 //Camera distance = 203035B0
 
-[Remove Blackbars]
-description=Removes black bars in cutscenes
-patch=1,EE,00190e58,word,24050000
\ No newline at end of file
+//[Remove Blackbars]
+//description=Removes black bars in cutscenes
+//patch=1,EE,00190e58,word,24050000
```
pgert, line 31: Why do you comment out this?

JordanTheToaster: Because both are broken.

crashGG:
> SPS

What does SPS mean?

F0bes:
Spiky Polygon Syndrome
Usually due to vertices that are clamped improperly, and they usually manifest as very extreme triangles.

Mrlinkwii:
@crashGG  Spikey Polygon Syndrome (SPS). SPS is the tendency of world geometry or character models to deform into wildly incorrect shapes,

crashGG:
Then, disc2 with the same elf checksum as disc1 should also have the same situation?

pgert:
> Then, disc2 with the same elf checksum as disc1 should also have the same situation?

@JordanTheToaster:
DMC 2 disc 2 still has the hack enabled - #557.
Any comment?

JordanTheToaster:
>Any comment?

That's classified.

PR #499, opened 2025-02-22
NFS Carbon: Correct the serial the game uses to detect Most Wanted Black Edition saves

CookiePLMonster:
Fixes an old bug where only base Most Wanted saves granted the cash bonus, and MW Black Edition saves did not. This is because the game used an incorrect serial to detect BE saves.

image

ProStreet is seemingly unaffected, as it detects Most Wanted, Most Wanted BE, Carbon, Carbon CE correctly.


CookiePLMonster:
Added support for `SLES-55006`.

Mrlinkwii (approved):

PR #501, opened 2025-02-24
Fixes to RB:C1GP [SLPM-65897] WS patch + pnach for English patched version

igorciz777:
Adds widescreen rendering to mirrors and reflections
patch

Also adds a pnach file for the English patched version (CRC 9C4C9611)
https://github.com/igorciz777/RBC1GP-Translation-Pipeline/releases/

### DIFF
--- a/patches/SLPM-65897_1C087362.pnach
+++ b/patches/SLPM-65897_1C087362.pnach
@@ -5,26 +5,26 @@ gsaspectratio=16:9
 author=wxvu & igorciz777
 
 //Rendering: General
-patch=1,EE,20167318,extended,3C043F10 //3C043F40
+patch=0,EE,20167318,extended,3C043F10 //3C043F40
 
 //Rendering: Reflections
-patch=1,EE,2016783C,extended,3C033F10 //3C033F40
+patch=0,EE,2016783C,extended,3C033F10 //3C033F40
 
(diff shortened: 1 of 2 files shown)


patches/SLPM-65897_1C087362.pnach
```diff
@@ -2,14 +2,31 @@ gametitle=Racing Battle C1 Grand Prix (J)(SLPM-65897)
...
 gsaspectratio=16:9
-author=github.com/igorciz777
+author=wxvu & igorciz777
 
-patch=1,EE,00167318,word,3c043f10 //3c043f40
+//Rendering: General
+patch=1,EE,20167318,extended,3C043F10 //3C043F40
 
+//Rendering: Reflections
+patch=1,EE,2016783C,extended,3C033F10 //3C033F40
```
CookiePLMonster, line 11: 1. The comment doesn't match the value patched.
2. Do they have to apply every frame? It's patching code so unless overlays are in use, you can just patch once. (this applies to all patches in these files)

igorciz777: 1. Comment refers to the original pre-patched value
2. Fixed

CookiePLMonster (approved):
Please squash when merging.

PR #506, opened 2025-03-06
Road Trip (USA) Patches (Progressive, widescreen, hi-quality cars, anti-blur, analog steering)

Souzooka:
Just a small patch file due to a request by silent, this converts an xdelta I made to a pnach which can be mix and matched without modifying the game.

This adds five switches:

1. No-Interlacing
Sets the game to use progressive mode upon boot; half vertical resolution

2. Force Hi-Quality Car Models
Cars in races/overworlds have pretty aggressive LODs, and this disables that. Additionally, this shows 3D wheel models on cars in the overworld when previously 3D wheels were never visible on cars in the overworld (to be specific, this makes the spokes parts of the wheels 3D, instead of a 2D texture).

3. Anti-Blur
This removes a ghosting effect in the game, which was implemented by taking the last two frames and overlaying it over the current frame with some transparency.

4. Analog steering
This one is more for fun; originally, the game only converts the analog into digital inputs. Implementation details are contained within the pnach file.

5. Widescreen 16:9
This is the meaty bit; this not only alters the aspect ratio of the 3D geometry, but aspect-corrects UI elements and repositions them to take advantage of the 16:9 screen space.

Pictures:
Road Trip_SLUS-20398_20250226150034
Road Trip_SLUS-20398_20250306012811
Road Trip_SLUS-20398_20250306012635
Road Trip_SLUS-20398_20250306012728
Road Trip_SLUS-20398_20250306012934
Road Trip_SLUS-20398_20250306012849
Road Trip_SLUS-20398_20250306013055
Road Trip_SLUS-20398_20250306013145
Road Trip_SLUS-20398_20250306013151

### DIFF
--- a/patches/SLUS-20398_4C168567.pnach
+++ b/patches/SLUS-20398_4C168567.pnach
@@ -1,11 +1,1262 @@
-gametitle=Road Trip (U)(SLUS-20398)
+gametitle=Road Trip (NTSC-U) (SLUS-20398)
 
 [Widescreen 16:9]
 gsaspectratio=16:9
-author=Arapapa
+author=Souzooka
+description=16:9 3D with 16:9-optimized UI
 
-//Widescreen hack 16:9
+// 3D aspect ratio
+patch=0,EE,2021F798,extended,3C013F40 // lui $at, 0x3F40
 
-patch=1,EE,0021f798,word,3c013f40 //3c013f80
+// Fix culling of 3D environment mesh subdivisions on sides of screen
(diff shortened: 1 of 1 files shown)


Souzooka:
Force-pushed to correct excess whitespace issues.

Souzooka:
Force-pushed to address feedback above. I tested the no-interlacing hack and I did not notice any interlacing artefacts with it active, so I replaced my own with it. I changed the `patch=1` to `patch=0` since it only writes to code so constantly writing is unnecessary.

Issue #508, opened 2025-03-07
GTA SA - SLUS-20946 - v1.03 frame pacing stutter due to patch


BParks21:
Using this versions widescreen patch causes weird very noticeable frame pacing stutter when enabled. Disabling the patch the game is buttery smooth. Just load this version of the game and pan the camera around slowly and consistently and it should easily be visible. Not much else to say. Patch on bad, patch off good.

MackdannyTwitch:
I'm having no issues on my end. There might be what appears to be a hiccup, but it also might just be assets loading and unloading. Its not really much of a hiccup, more like a nonosecond of something. In any case, the fps doesnt drop no matter what. Even if i also add the 60fps cheat on. At first i tried with just the WS and then with no WS. Didnt really see anything that would cause concern. Then again, it may just be a difference of hardware. Sorry I couldn't be much help.

BParks21:
Well to me it is not subtle at all. It is very obvious and bad. All I do is walk outside of CJ's House stand outside of the Street in front of the surrounding homes and slowly and consistently pan the camera. The stutters are very strong to me. As soon as I disable the patch and repeat the same steps the stutter is completely gone. I don't need any numbers or statistics to see it at all. I perceive it as bad pacing. The camera panning will appear smooth for a few seconds then it jerks. Then returns to normal for a very brief period. Back and forth.

BParks21:
@PeterDelta Thanks for fixing this it looks butter smooth now.

PR #509, opened 2025-03-07
Small fix

PeterDelta:
Better performance
#508

### DIFF
--- a/patches/SLUS-21560_D2790A77.pnach
+++ b/patches/SLUS-21560_D2790A77.pnach
@@ -1,14 +1,14 @@
-gametitle=Family Guy (SLUS-20718)
+gametitle=Family Guy - Video Game (SLUS-21560)
 
-[Widescreen 16:9]
-gsaspectratio=16:9
-author=Brandondorf9999
-description=Widescreen update
-patch=1,EE,0011B3A8,word,00000000 //10400009
-patch=1,EE,0011B3BC,word,3C013F80 //3C013F59
-patch=1,EE,0011B3C0,word,34210000 //3421999A
+//[Widescreen 16:9]
+//gsaspectratio=16:9
+//author=Brandondorf9999
+//description=Widescreen update. Breaks eye textures on models.
+//patch=1,EE,0011B3A8,word,00000000 //10400009
+//patch=1,EE,0011B3BC,word,3C013F80 //3C013F59
+//patch=1,EE,0011B3C0,word,34210000 //3421999A
 
-[60 FPS]
-author=asasega
-description=Patches the game to run at 60 FPS.
-patch=1,EE,20259698,word,00000001 //00000002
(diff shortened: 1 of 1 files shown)


MackdannyTwitch:
I take it you fixed that issue from.thst bug report?

CookiePLMonster:
In this patch, multiple extended writes (including the one you fixed) are 1-byte (`0` as the first address digit), but the immediate provided is 4-byte. Is this intentional?

CookiePLMonster:
> To further satisfy this curiosity, the reason I write it like this is because it is easier for me to convert it to raw and test it on the ps2.

Can't raw ps2 patches do 1/2 byte writes?

refractionpcsx2:
What Cookie is asking, is it looks like the line has been converted from `word` to `extended` without modifying anything else. it is currently set to 8 bit mode (by the 0 prefix) but you left in a 32bit word as a value, instead of just having `AC`, which is all you need.

refractionpcsx2:
I think for clarity it should be changed but you can use // to put a comment after it to explain what you're doing, if you feel it necessary :)

patches/SLES-52541_A1B3F232.pnach
```diff
@@ -4,16 +4,17 @@ gametitle=Grand Theft Auto - San Andreas (PAL-M5) (v1.03) SLES-52541 A1B3F232
...
+patch=1,EE,0020A498,extended,AC
+patch=1,EE,0020817C,extended,AC
+patch=1,EE,002ECAB0,extended,40
+patch=1,EE,002ECB00,extended,D0
 patch=1,EE,0033DAB4,extended,3C013F80 //scenes zoom
-patch=1,EE,E0010001,extended,0066C17C
+patch=1,EE,E0020001,extended,0066C17C
 patch=1,EE,0033DAB4,extended,3C013F40
 patch=1,EE,2021EDB4,extended,3C023FC0
+patch=1,EE,2021EDB4,extended,3C024000
```
CookiePLMonster, line 17: What happened here?

CookiePLMonster: Ah, right, the write above it is conditional. Do you know the exact function you're patching here? If not I can look it up.

CookiePLMonster: I mean the game function, not an instruction.

CookiePLMonster: OK I checked, the function that is patched here is `CCamera::CalculateFrustumPlanes` and it's patching the `2.0` divisor of:
```c
val = cos((0.017452778 * CDraw::ms_fFOV) / 2.0);
```

Essentially this change makes the game cull less entities, so the performance hit makes sense.

MackdannyTwitch:
Heh, I was wondering what happened.

PR #512, opened 2025-03-09
Add Air Ranger 2 Plus anti-blur patch

Souzooka:
Gives the player glasses so that they can properly fly a helicopter.

Must be enabled outside a mission to properly stop the depth of field task from spawning.

Before:
Air Ranger 2 Plus - Rescue Helicopter_SLPM-65409_20250309170122

After:
Air Ranger 2 Plus - Rescue Helicopter_SLPM-65409_20250309170747_(2)


Gabominated:
Now it's looks very nice.

PR #514, opened 2025-03-09
Add 60fps hack for Skyscraper

Souzooka:
Converts the video mode to NTSC, which the game handles gracefully (besides the one caveat where dialogue gets cut off early inside cutscenes) and allows for 60fps.


Gabominated:
hello, well, since we have been asked to make separate patches when possible, I suggest that as PerterDelta says, it be renamed NTSC Mode and your fps line be removed because if you separate them, then there would be 2 fps patches and if they have the same behavior as you say then there is no point in having 2 FPS patches, all that would be left would be to rename the patch above also to 50/60 FPS.

PR #521, opened 2025-03-23
Gran Turismo addiontal patches & God Hand Dead Pan

SilverKeeper:
GT4 Prologue .pnach added with some patches, added additional patches to GT3 and GT4, and added the Dead Pan restoration code for God Hand (the code is the same for NTSC and PAL).


CookiePLMonster:
I don't see the point in submitting patches that should be tweaked by the user to their liking, like the remappable controls or customizable units.

pgert:
SLUS-21503_6FB69282:

The '[No-Interlacing]' hack was made by umechan:
https://forums.pcsx2.net/Thread-No-interlacing-codes?pid=606681#pid606681
So if you're up for it, add "author=umechan" to the hack.

F0bes:
Closing due to inactivity.

PR #529, opened 2025-03-29
Add de-interlace and ntsc mode patches for Project Minerva Professional (EU)

Souzooka:
Adds patches to remove interlacing artefacts and force the game to NTSC upon startup (after reaching the title menu, you can scroll down to options and change to NTSC, but this patch just does that for the user).

### DIFF
--- a/patches/SLES-52472_0AF4683D.pnach
+++ b/patches/SLES-52472_0AF4683D.pnach
@@ -1,0 +1,15 @@
+gametitle=Project Minerva Professional (PAL) (SLES_524.72)
+
+[No-Interlacing]
+author=Souzooka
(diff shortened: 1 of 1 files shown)


pgert:
Any comment on the hack by raceywacey?
https://forums.pcsx2.net/Thread-No-interlacing-codes?pid=641813#pid641813

patch=1,EE,201AACCC,extended,00000000
patch=1,EE,201C21EC,extended,00000000

Souzooka:
I went and found 1C21EC myself, 0x1AACCC here seems to be a false positive -- it isn't being hit during in-mission gameplay/menus/FMVs from what I can see.

PR #530, opened 2025-03-29
Update patches

PeterDelta:
Updated patches, relevant changes:

- God of War: Updated the widescreen patch to fix the menu flames.
- X-Men 2: Disabled the widescreen patch because it causes rendering errors and crashes after level 3. I haven't been able to get a proper widescreen version.
- Reservoir dogs: Removed 60Hz patch in PAL version, FMVs are not fully synchronized. Fixed frame rate in FPS patch, now it's accurate.

The rest have been tested without issue.

### DIFF
--- a/patches/SCES-51164_A2837592.pnach
+++ b/patches/SCES-51164_A2837592.pnach
@@ -70,3 +70,8 @@ patch=1,EE,0029F904,word,31310000 //31310001
 patch=1,EE,002A1CAC,word,30B20000 //30B20001
 patch=1,EE,00326DBC,word,00000000 //64420008
 patch=1,EE,00326E44,word,00000000 //64420008
+
+[NTSC Mode]
+author=PeterDelta
+description=NTSC mode at start.
+patch=0,EE,0010024C,word,240A0002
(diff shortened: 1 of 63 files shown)


bigol83:
I think this is related to the issue i'm having with the 480p patch on God of War. The very beginning section of the boat battle disengages the progressive scan and the game goes back to 50 fps instead of 60. Also, when this happens the game goes like in slight slow motion. It fixes itself after the first fmv plays when the other soldiers appear. 

https://github.com/user-attachments/assets/5e1a5b48-3a8a-46cb-a13c-dd46328bce3a

Also, the green orbs coming from the chest have that black spot that is not there without the progressive scan patch. I checked to be sure it was not something with upscaling but it isn't.

immagine

BParks21:
Curious, are there any issues with GoW 2's WS patch? Or is that one fine?

pgert:
As pointed out earlier (by someone else in some other PR),
 leaving the last line of the pnach blank simplifies PR presentation.

Mrlinkwii:
may i ask whats the status of this ?

PR #531, opened 2025-03-30
Allowing the patches for some games to work on the Retroachievements …

MackdannyTwitch:
…supported counterparts

Jak and Daxter, Jak 2, and Max Payne have Patches that, while work, are not recognized by their Retroachievements enabled counterparts.

The reason for this is for the Jak games, an Xdelta patch was made to disable the debug menu, which can make it possible to cheat. For Max Payne, a patch was made to deter cheating as well.

Ultimately, in both scenarios, the CRC was changed, so this submission recognizes those CRC changes. No other changes to the base game experiences were made, and the PNACH files submitted are otherwise identical to the base game's PNACH files, but confirmed to work on the altered CRC.


MackdannyTwitch:
Ready for review.

PR #533, opened 2025-03-31
Add Simpsons Skateboarding 60fps patch

Souzooka:
Does what it says on the tin -- there's a boolean set to 1 which tells the game to implement a framerate cap, and this patch makes the game treat that boolean as if it is always 0.


Souzooka:
There's a timestep in milliseconds at 0x3343D0, though lowering this value causes the game to speed up.

PR #534, opened 2025-03-31
Add Simpsons Skateboarding 60fps patch

Souzooka:
This is a modification of the patch for the previous pull request which only uncapped the framerate, but did not touch timestep: https://github.com/PCSX2/pcsx2_patches/pull/533

Timestep (and game speeds calculated from it) are only calculated on boot, so this patch manually recreates each of those calculated values set in code.

### DIFF
--- a/patches/SLUS-20114_6D0D9F1F.pnach
+++ b/patches/SLUS-20114_6D0D9F1F.pnach
@@ -1,0 +1,39 @@
+gametitle=Simpsons Skateboarding, The (NTSC-U) (SLUS-20114)
+
+[60 FPS]
+author=Souzooka
+description=Removes framerate cap.  Simulation runs at 60fps, but some elements such as UI run quick.
+
+// First of all, only patch if we're in the game ELF to avoid bugs
+// (the FMV player is a separate ELF)
+patch=1,EE,E00EFFF0,extended,00100FF0 // check 00100FF0 matches value xxxxFFF0 (arbitrary code address)
+
+// Removes framerate cap
+patch=1,EE,20508D18,extended,0 // 1
+
+// Sets timestep to 16ms (from 33ms) so the game updates 60 times a second.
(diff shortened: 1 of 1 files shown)


Gabominated:
Hello, great work, looks good, only one thing, the turning is double speed.
https://github.com/user-attachments/assets/63465f30-ff88-40cf-86e1-655742027e7b

Souzooka:
Added some lines which affect stationary turn speed (there is different code for handling turning while moving and while stationary, and turning while moving seems to be more or less the same on 30 or 60 FPS).

Souzooka:
Also added a patch onto the end which disables Kent's VO about your tricks in game.

Souzooka:
Scratch that, there's a game option for it -- gonna stop messing with this game right now, hah

MackdannyTwitch:
Nothing for WS?

MackdannyTwitch:
Ah yeah, I should have phrased that differently. That is my bad. I'm just bewildered thwres nothing for WS for that game. Not even a cancelled out code. Good work has been done, nonetheless. A good 60fps code is always welcome.

Gabominated:
Are you able to make a port to PAL version?

Souzooka:
Aren't there 5 PAL versions?

MackdannyTwitch:
Shoot, you weren't kidding.


Screenshot_20250401_144956_Chrome

Gabominated:
Yeah, I think that all PAL version can use one single code, I have a NTSC patch to PAL-S in my repository and I can make a test in this version.

Souzooka:
Doesn't seem like variables match up on each PAL version, looking at this TCRF article https://tcrf.net/The_Simpsons_Skateboarding#--60 different variable addresses are set, and it'll probably be similar for editing the other variables and code.

Gabominated (approved):

PR #539, opened 2025-04-07
Drive to Survive: Remove a pointless widescreen patch

CookiePLMonster:
The original release of **Mashed** (SLES-52446) did not have native widescreen support, but updated releases **Mashed: Fully Loaded** (SLES-53152) and **Drive to Survive** (SLUS-21109) have native WS available from the game options. This patch does the same thing visually, and stacks on top of the native support.

In-game Widescreen Off, patch enabled:
image

In-game Widescreen On, patch enabled:
image

In-game Widescreen On, patch **disabled**:
image

### DIFF
--- a/patches/SLUS-21109_0520A26D.pnach
+++ b/patches/SLUS-21109_0520A26D.pnach
@@ -1,18 +1,20 @@
 gametitle=Drive To Survive (U)(SLUS-21109)
 
-[Widescreen 16:9]
-gsaspectratio=16:9
-author=Arapapa
+// Mashed: Fully Loaded (SLES-53152) and Drive to Survive (SLUS-21109) have native widescreen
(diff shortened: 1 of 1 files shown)


Mrlinkwii (approved):

MackdannyTwitch:
Out of curiosity, i take it the patch doesn't enable the widescreen, as you were able to have the same results even with the in-game option off?

I always thought auto enabling/locking patches were cool, but if it's as you ssy, then yeah, i get you.

CookiePLMonster:
> Out of curiosity, i take it the patch doesn't enable the widescreen, as you were able to have the same results even with the in-game option off?

Yeah, for some reason it reimplements widescreen instead of toggling the option. If it did that, I wouldn't have touched the patch at all as that's fine.

PR #541, opened 2025-04-10
Improved Rayman Revolution (NTSC-U) Widescreen Patch

RibShark:
Fixes all 2D elements, including light coronas. Tested through the entire game with 100% to make sure everything is working. Can be adapted for other aspect ratios by modifying the first line (examples are given in comments).

PR #545, opened 2025-04-20
Rayman 2 - porting and formating

pgert:
The NTSC-K disc seems to be identical to the NTSC-U disc.
Not only do they have the same CRC, but they had the same WS hack by ElHecht & icup321.
Consequently it should be resonably safe to apply the new WS hack
 by RibShark (Rib) from #541 for the NTSC-K disc as well.
Testing & feedback appreciated.

Also formated the hack for both discs, and re-added the old hack (though commented out) to avoid confusion.
@RibShark - can you test the formating (with 21:9)?


RibShark:
Can confirm the new formatting works great and that the additional patches (tested with 21:9) work correctly. Korean version also works correctly.

Mrlinkwii (approved):
Assumed tested and working

F0bes (approved):

PR #546, opened 2025-04-23
Add 60fps 007 Quantum of Solace US patch

Souzooka:
This patch adds 60fps functionality to the US version of 007 - Quantum of Solace. The patch conditionally switches to 30fps when .pss files run, to prevent them running too quickly. It also affects the weapon recoil compensation, which seems to be affected by framerate, to recreate similar recoil behavior as if the game was running at 30fps.

### DIFF
--- a/patches/SLUS-21813_5A6A935D.pnach
+++ b/patches/SLUS-21813_5A6A935D.pnach
@@ -1,0 +1,24 @@
+gametitle=007 - Quantum Of Solace (NTSC-U) (SLUS-21813)
+
+[60 FPS]
+author=Souzooka and asasega
+description=60 FPS; recommend using 180% EE Cycle Overclock
+
+// Framerate
+patch=1,EE,203CA748,extended,3C // 60 FPS
+patch=1,EE,E0010001,extended,003DDB84 // Check 003DDB84 is XXXX0001, indicating a .pss is playing
+patch=1,EE,203CA748,extended,1E // 30 FPS (normal .pss speed)
+patch=1,EE,E0010000,extended,003CAA98 // Check 003CAA98 is XXXX0000, workaround for softlock after subduing first opera house enemy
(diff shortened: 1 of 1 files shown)


Mrlinkwii (approved):

Souzooka:
There seems to be a softlock when subduing the first enemy in the opera house level at 60fps. I'm going to investigate a workaround to this.

Gabominated:
This patch uses the same address and value as asasega at the 60fps forum.
[https://forums.pcsx2.net/Thread-60-fps-codes?pid=550658#pid550658](url)

Souzooka:
I've put in a workaround for the softlock after subduing the first enemy at the Austrian Opera House, and played up from a new game to that point to make sure there are no issues. Still testing the rest of the game.

@Gabominated would you suggest any changes? I found the address on my own, and the game wouldn't even be complete-able with just that one line due to the mentioned softlock (not to mention pre-rendered cutscenes being twice as fast and recoil being off).

Souzooka:
I added a credit to asasega; I also added in some more logic that circumvents another cutscene softlock I discovered. I played through the rest of the game and didn't encounter any more softlocks, so the patch should be fully playable and behave like 30fps.

patches/SLUS-21813_5A6A935D.pnach
```diff
@@ -0,0 +1,20 @@
+gametitle=007 - Quantum Of Solace (NTSC-U) (SLUS-20932)
```
pgert, line 1: Wrong serial-code: "SLUS-20932" -> "SLUS-21813".

pgert:
Suggestion:

[Onscreen fps Display]
author=asasega
patch=1,EE,003CD06E,word,00000001

Souzooka:
Fixed the game serial and added the FPS display patch.

Gabominated (approved):

PR #549, opened 2025-05-06
Raw Danger (PAL) - Updates

MrWrightyLC:
**Raw Danger  (Regular version)**
- Updated gametitle
- Add Remove Colorgrading Filter patch

**Raw Danger  (Undub & NTSC Mode versions)**
- Updated gametitle (Undub & NTSC Mode shares the same CRC)
- Add Remove Colorgrading Filter patch

**Raw Danger  (NTSC Mode + Performance version)**
- Updated gametitle
- (Unnecessary to add Remove Colorgrading Filter patch here, since romhack already includes it.)

Some screenshots before and after with the Remove colorgrading filter patch
Raw Danger_SLES-54587_20250505170011
Raw Danger_SLES-54587_20250505170336
Raw Danger_SLES-54587_20250505173428
Raw Danger_SLES-54587_20250505173536
Raw Danger_SLES-54587_20250505174302
Raw Danger_SLES-54587_20250505174322


JordanTheToaster:
Please remove all of the URLs you have added.

Souzooka:
Hi, I've also located the addresses for the US and JP versions of the game where a similar effect can be achieved:

US= 0x228430
JP= 0x2279D0

This is where the same function as is in PAL at 0x2283C0 exists in those versions, so the colorgrading effect can be removed by having those two lines there. If you can add those that'd be pretty neat.

F0bes (approved):

PR #552, opened 2025-05-07
GTA San Andreas: revert values original in cutscenes

### LINKED ISSUE #551: GTA San Andreas: FOV issues
1. Huge FOV in interiors results in objects vanishing on the edges often.
   <details><summary>Notice the dishes on the table on the right</summary>

   Image
   Image

   </details>

2, @PeterDelta Your FOV changes revert when widescreen bars are enabled from the script, but it's not extended to mocap cutscenes. This results in many cutscenes showing unintended elements offscreen.
    <details><summary>T-posing Big Smoke in the bottom left corner</summary>
   
Image

   
   </details>

Do you know exactly what variable are you checking for? I might be able to help finding a more reliable variable - or at least a second one to check in an OR-like manner.

PeterDelta:
We've reverted the cutscene improvements to their original values. I'll take this opportunity to add the PAL-G version.

Fixes #551 

Note: Improved the "remove ghosting affects" patch on PAL versions.

### DIFF
--- a/patches/SLES-52541_B440A8FE.pnach
+++ b/patches/SLES-52541_B440A8FE.pnach
@@ -10,11 +10,6 @@ patch=1,EE,0020A4F8,extended,AC
 patch=1,EE,002081DC,extended,AC
 patch=1,EE,002ECBE0,extended,40
 patch=1,EE,002ECC30,extended,D0
-patch=1,EE,0033DC44,extended,3C013F80 //scenes zoom
-patch=1,EE,E0020001,extended,0066C8FC
-patch=1,EE,0033DC44,extended,3C013F40
-patch=1,EE,2021EE14,extended,3C023FC0
-patch=1,EE,2021EE14,extended,3C024000
 patch=1,EE,0026E18C,word,3C084280
 patch=1,EE,00269678,word,3C024280
 patch=1,EE,0026E7A8,word,3C024280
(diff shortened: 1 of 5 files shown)


CookiePLMonster:
Doesn't this undo the FOV changes in gameplay entirely?

CookiePLMonster (approved):
I've been playing with this version of the patch since it was submitted and haven't noticed any more issues like this any longer. Seems to be all good now.

patches/SLES-52541_B440A8FE.pnach
```diff
@@ -72,9 +67,7 @@ patch=1,EE,002A8A1C,word,E42D2FA8
 author=PeterDelta
 description=Removes the ghosting effect keeping radiosity.
 patch=1,EE,00668D5C,extended,00000000
-patch=1,EE,00668DE8,extended,00000000
-patch=1,EE,E0010100,extended,007005E0
-patch=1,EE,007005E0,extended,00000130
+patch=1,EE,00668DE4,extended,00000001
```
CookiePLMonster, line 70: Submitted by mistake? If not, why 1 and not 0?

CookiePLMonster:
> I was going to mention this, but you got ahead of yourself xD. It's subtle, let's say it's a level. The original value of 2 creates ghosting because it overlaps the image; a value of 1 adjusts it to the contour, and a value of 0 overlaps it below. A value of 1 is the most balanced.

This might depend on the scaling factor. There are also `CPostEffects::m_RadiosityFilterUCorrection` and `CPostEffects::m_RadiosityFilterVCorrection` values right below `CPostEffects::m_RadiosityFilterPasses` that gets corrected here, I'll have to check if they are used. It might be that changing those instead of the filter passes is **the** best approach.

CookiePLMonster:
I re-tested with the maximum scaling PCSX2 allows and with 1 filter pass ghosting is definitely still here, while with 0 it's completely gone. Are you sure you've not accidentally changed the UV correction values in one of the executables, as they are also `2`? I tried changing those to `0` and I indeed observed a similar effect where it overlapped in the other direction instead.

Take a look here:

filterPasses = 2
filterPasses2

filterPasses = 1
filterPasses1

filterPasses = 0
filterPasses0

It seems clear to me that nothing overlays with 0 passes, the effect is completely gone.

kamfretoz (approved):

F0bes (approved):

PR #554, opened 2025-05-10
GTA San Andreas: Changes to postfx removal patches

CookiePLMonster:
# Description

This PR introduces two changes to GTA San Andreas patches:
1. `Remove Radiosity Filter` patch in fact broke all the post effects, including the color filter, radiosity, heat haze, night and thermal vision, motion blur, and the underwater ripple effect. We have better alternatives now, so it makes no sense to keep a cheat that breaks the game.
   It's possible, however, that some people were using this patch to remove the yellow tint and make the game look more like the PC port. For this reason, I introduced a new `Remove Color Filter` patch that does the removal properly. The new patch **doesn't** remove the ghosting effect caused by radiosity, but it can safely be used with the other patch to also remove that, if users desire so.
2. The current `Remove Ghosting Effects` patch from @PeterDelta is near perfect, as it removes the ghosting effect without impacting the other post effects. However, it removes radiosity by removing its render passes entirely, which very slightly washes out the colors as radiosity normally produces a slight bloom effect. I introduced a new patch `Remove Radiosity Filter Pass` that keeps the render passes, but removes the filter passes, thus removing the offset. This preserves the colors perfectly, but removes the blurriness.

# Screenshots

All screens are done with 4x scaling. Best to open in a new tab and switch back and forth to see the difference.

<details>
<summary>Stock</summary>

Grand Theft Auto - San Andreas_SLUS-20946_20250510130237
Grand Theft Auto - San Andreas_SLUS-20946_20250510130303
Grand Theft Auto - San Andreas_SLUS-20946_20250510130317

</details>

<details>
<summary>PeterDelta's radiosity removal - sharp, but with colors slightly washed out</summary>

Grand Theft Auto - San Andreas_SLUS-20946_20250510130639
Grand Theft Auto - San Andreas_SLUS-20946_20250510130716
Grand Theft Auto - San Andreas_SLUS-20946_20250510130802

</details>

<details>
<summary>My radiosity removal - sharp and with colors intact</summary>

Grand Theft Auto - San Andreas_SLUS-20946_20250510131204
Grand Theft Auto - San Andreas_SLUS-20946_20250510131555
Grand Theft Auto - San Andreas_SLUS-20946_20250510131636

</details>

<details>
<summary>Color filter removal</summary>

Grand Theft Auto - San Andreas_SLUS-20946_20250510132741
Grand Theft Auto - San Andreas_SLUS-20946_20250510132759
Grand Theft Auto - San Andreas_SLUS-20946_20250510132817

</details>

# To Do

- [x] Test the patch more.
- [x] Figure out if `Remove Ghosting Effects` should be cut.
- [x] Port to other versions.

### DIFF
--- a/patches/SLES-52541_A1B3F232.pnach
+++ b/patches/SLES-52541_A1B3F232.pnach
@@ -1,4 +1,4 @@
-gametitle=Grand Theft Auto - San Andreas (PAL-M5) (v1.03) SLES-52541 A1B3F232
+gametitle=Grand Theft Auto: San Andreas (PAL-M5) (v1.03) SLES-52541 A1B3F232
 
 [Widescreen 16:9]
 gsaspectratio=16:9
@@ -63,21 +63,24 @@ patch=1,EE,002A8904,word,3C01007C
 patch=1,EE,002A8908,word,03E00008
 patch=1,EE,002A890C,word,E42D2418
 
-[Remove ghosting effects]
-author=PeterDelta
-description=Removes the ghosting effect keeping radiosity.
-patch=1,EE,006685DC,extended,00
-patch=1,EE,00668664,extended,00
-patch=1,EE,006686A8,extended,00
-patch=1,EE,006686AC,extended,00
-patch=1,EE,006686B0,extended,00
-patch=1,EE,006686B4,extended,00
+[Remove Ghosting Effects]
+author=Silent, PeterDelta
+description=Removes the ghosting effect from radiosity and color filter post effects, preserving the slight bloom effect. Also disables the seam remover post effect.
 
+patch=0,EE,006685DC,extended,00 // CPostEffects::m_bSeamRemover
+patch=0,EE,20668664,extended,00000000 // CPostEffects::m_RadiosityFilterPasses
 
-[Remove Radiosity Filter]
-author=refractionpcsx2
-description=Removes the radiosity filter which causes a ghosting effect on the people and environment.
-patch=1,EE,2051A0C8,extended,00000000
+// These values are read from stream.ini from the init overlay, so we have to patch them every frame
+patch=1,EE,206686A8,extended,00000000 // CPostEffects::m_colourLeftUOffset
+patch=1,EE,206686AC,extended,00000000 // CPostEffects::m_colourRightUOffset
+patch=1,EE,206686B0,extended,00000000 // CPostEffects::m_colourTopVOffset
+patch=1,EE,206686B4,extended,00000000 // CPostEffects::m_colourBottomVOffset
+
+[Remove Color Filter]
+author=Silent
+description=Removes the color filter, making the game look more like the PC version, without the yellow tint.
+
+patch=0,EE,20515658,extended,00000000 // NOP CPostEffects::ColourFilter
 
 [50 FPS]
 author=Boludoz
--- a/patches/SLES-52541_B440A8FE.pnach
+++ b/patches/SLES-52541_B440A8FE.pnach
@@ -1,4 +1,4 @@
-gametitle=Grand Theft Auto - San Andreas (PAL-M5) (v2.01) SLES-52541 B440A8FE
+gametitle=Grand Theft Auto: San Andreas (PAL-M5) (v2.01) SLES-52541 B440A8FE
 
 [Widescreen 16:9]
 gsaspectratio=16:9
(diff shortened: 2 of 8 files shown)


CookiePLMonster:
@PeterDelta What do you think should be done with your existing patches? Do you think it makes sense to keep both radiosity removal patches in, or should mine replace yours as it's smaller and doesn't affect the color correction? We could do one of the following:

1. Keep your patch as-is, and add mine.
2. Replace your patchwith mine.
3. Keep both patch, but replace yours with mine under the current `Remove Ghosting Effects` name, so people currently using the patch will automatically "upgrade", but can still use yours if they wish to.

Gabominated (approved):

MackdannyTwitch:
Is this issue only present in the launch version? I wish i was good at porting, I'd port it to 3.00.

CookiePLMonster:
> Is this issue only present in the launch version? I wish i was good at porting, I'd port it to 3.00.

No, all versions have the same postfx. I simply didn't get to porting it yet, that's why it's a draft.

Gabominated:
The truth is that these patches make the image look very good but sometimes you have to pay a very high price and the game ends up broken, it can be understood why many look for this kind of patches but the truth is that they need a lot of testing to determine that it doesn't break something important in the game, I can help with port it if these patches pass the tests.

CookiePLMonster:
> The truth is that these patches make the image look very good but sometimes you have to pay a very high price and the game ends up broken

??? We literally have DWARF symbols for San Andreas, so function names and parameter names are known, and with SkyGfx on PC the entire post-effects pipeline of San Andreas is well understood. There is no randomness involved here.

Gabominated:
> > The truth is that these patches make the image look very good but sometimes you have to pay a very high price and the game ends up broken
> 
> ??? We literally have DWARF symbols for San Andreas, so function names and parameter names are known, and with SkyGfx on PC the entire post-effects pipeline of San Andreas is well understood. There is no randomness involved here.
I'm referring to patches in general that make things look better, cleaner and sharper. I personally love them. It's just a slightly rambling comment. I only offer myself as a porter if needed.

Gabominated:
> > The truth is that these patches make the image look very good but sometimes you have to pay a very high price and the game ends up broken
> 
> ??? We literally have DWARF symbols for San Andreas, so function names and parameter names are known, and with SkyGfx on PC the entire post-effects pipeline of San Andreas is well understood. There is no randomness involved here.

I'm referring to patches in general that make things look better, cleaner and sharper. I personally love them. It's just a slightly rambling comment. I only offer myself as a porter if needed.

CookiePLMonster:
> By the way, this address that I included in the "remove Ghosting Effects" patch what it does is also eliminate that duplicating effect of the objects making it look sharper, I would also leave it within it since it visually complements it:
> patch=1,EE,006684DC,extended,00000000

Fair enough. I don't have the symbols for that particular effect since it's not on PC or mobile, but I do know it's a post effect for sure, and therefore likely safe to disable. I'll update the PR restoring it, just making it an one-time patch as that boolean resides in a read-only area of the module.

CookiePLMonster:
I enabled the seam search mode and immediately started noticing gaps - this is at 4x scaling, but with higher upscaling it might be even worse because sub-pixel seams now may span multiple pixels:

Grand Theft Auto - San Andreas_SLUS-20946_20250511105307

Considering how popular upscaling is, I'm not sure disabling this filter is a good idea.

CookiePLMonster:
I replaced the old patch with the new one, and re-introduced the removal of the seam remover, just as an one-time patch. I'll see how that looks when I play the game later this weekend.

If everything is OK, I'll port to the other executables and then it can be undrafted.

CookiePLMonster:
Surprise! @aap reminded me that the color filter offsets itself by half a pixel (as per the values read from `stream.ini`), and we can undo that to make the image even sharper.

Before:
Grand Theft Auto - San Andreas_SLUS-20946_20250511121809

After:
Grand Theft Auto - San Andreas_SLUS-20946_20250511121840

The change is subtle, but CJ's tank top, as well as the wall to the left is noticeably sharper when the offset is removed.

CookiePLMonster:
I'll keep this patch "in testing" this weekend, then proceed to porting it early next week so it can be un-drafted and merged. I should be able to port it to all game executables, even the German ones.

PR #569, opened 2025-06-02
ws: Digital Devil Saga

RedDevilus:
Fix patch so people can use widescreen patch (also in hardcore retroachievements)


RedDevilus:
Seems to work fine, but can be ported to other serials if needed.

PR #571, opened 2025-06-05
Add framerate hack for Vexx (NTSC-U)

FrankyBuster:
Adds framerate hack to default the framerate at 60fps, so it no longer has the wait time for the cap to adjust after overclock.


Gabominated:
But this game already run at 60fps, there is something that I miss?

FrankyBuster:
>But this game already run at 60fps, there is something that I miss?
﻿
I've mentioned this before, the game has a dynamic framerate cap that always defaults to 30 each time it loads a level/sub level and takes a few seconds to change to 60, this code makes the cap default to 60 at all times for consistency so it better matches with the better running gamecube and xbox versions.

PR #585, opened 2025-07-15
A Different Widescreen Patch for Ratchet & Clank (NTSC-U)

ghost:
I found a different widescreen patch that appears to have been made by someone by the name of **TechieSaru**. I did not create the patch.

People have stated that the old patch has an issue but on the multiple systems I tested the patch with, I did not notice any issue that justifies the old patch not being included with PCSX2. The only notable issues I noticed were with the menu when returning to the main menu after loading a save file as well as certain elements not being adjusted for the 16:9 aspect ratio. Even though people have stated there is a reason why the old patch is not included with PCSX2, I have not been provided any evidence of an issue that justifies the old patch not being included with PCSX2 as of recently.

Anyway, this patch appears to be different and also appears to fix the issue with the menu when returning to the main menu after loading a save file. Therefore, this patch appears to be an improvement compared to the old patch.

The person by the name of **TechieSaru** stated the following on a website post:

"NOTE: This archive contains PCSX2 patches that are currently not in the PCSX2 community archive. Do NOT request to have them added to the archive. They are being posted independently for a reason.

If you find any issues with my patches or have any requests, feel free to contact me via discord."

I do not know what was meant by "PCSX2 community archive". This repository appears to be an official repository for PCSX2 so the person might have been referring to something different but I do not know. The statement appears to be referring to the PCSX2 patches in general that are provided in the post, which are for multiple games and are not just widescreen patches. I do not know what the reason was or how the statement should be interpreted but the PCSX2 team can decide what to do with the information provided and possibly correctly interpret what was meant by the statement. I am simply trying to provide information as it was found. If **TechieSaru** can be contacted, I would recommend that the PCSX2 team do so to clarify what was meant by the statement and if the person is okay with this patch being added to this repository.

There seem to be patches for multiple games that appear to be made by **TechieSaru** that are in this repository.

Later in the post, the following is also mentioned:

"Ratchet & Clank (NTSC-U) - Re-worked widescreen patch. My old patch had issues I was not made aware of. Now includes an FMV's fix."

According to the website for downloading the patch, it appears that the patch was possibly modified as recently as June 4th, 2025.


sandboxgames123:
Okay. Like I mentioned, I do not know what "PCSX2 community archive" refers to. If the PCSX2 team believes it refers to this repository (or collections like this repository) then I would agree with not including it unless the creator approves it. I would recommend the PCSX2 team to attempt contacting **TechieSaru** and ask if the person is okay with the patch being included.

JordanTheToaster:
> I understand that the author doesn't want it published here

I would personally like the reason why as if this does fix the issues present in older patches we should include.

PR #588, opened 2025-07-25
WS: Disable project zero PAL WS

JordanTheToaster:
Appears to break mirror rendering.

With:
image

Without WS:
image


pgert:
Does this occur in gameplay and/or cinematics?

PR #592, opened 2025-07-29
SLUS-20443: Correct gravity at 60fps (Rugrats: Royal Ransom)

Souzooka:
Amends the 60fps code for the US version of Rugrats: Royal Ransom. When changing the framerate divisor alone, gravity is too aggressive and the player cannot vertically jump nearly as high as in 30fps. This patch changes that behavior to approximate the correct jump height/gravity at 60fps.


Souzooka:
Looking into a cleaner way of implementing this -- as it is unfortunately at the moment in-game slowdowns cause a moon-gravity-like effect. The issue is that function 0x00119DF8 processes gravity in 1/60th of a second ticks (representing by doing a gravity pass, adding 0.0166667f to a sum, then checking if that sum is equal to or greater than the provided delta time), but is provided a delta time rounded to the nearest thousandth.

This means that at 30fps, the function is provided a deltatime of 0.33f.
0.0166667 + 0.0166667 >= 0.033f ; two gravity iterations

However, at 60fps, the function is provided a deltatime of 0.17f.
0.0166667 < 0.017f
0.0166667 + 0.0166667f >= 0.017f ; still two gravity iterations!!

This effectively causes the player's vertical velocity to decelerate twice as fast as intended. It's not particularly an easy fix, because as is each time gravity is processed it is completely frame independent. So, at 60fps, if only one gravity iteration occurs, at 59fps, every single frame, two gravity iterations occur, and the player drops to earth much faster than intended.

I want to take another look at this and perhaps hook the beginnings and end of this function to perhaps roll over the remainder of the time left to the next function call. So for example, after doing two gravity iterations on the first frame at 60fps, the sum starts at 0.0163334f (0.0166667f + 0.0166667f - 0.017f) on the next frame, meaning it'll only do the one iteration then exit, as 0.0163334f + 0.0166667f >= 0.017f. This should also make jump heights more consistent when the game is running at a framerate that is not an exact divisor of 60.

Souzooka:
I should note that technically the rollover time is updated every time the gravity function is called, but I believe it is only called for the player character, so it should be functionally the same as only updating it once per frame.

Mrlinkwii (approved):

PR #595, opened 2025-07-30
Fix datetime issue when using computers (Resident Evil: Code Veronica)

VelpaChallenger:
Whenever you use a computer, either as Chris or Claire, the time shown is in Japan timezone, instead of local. This patch fixes that and allows you to see the time (presumably) as originally intended.

### DIFF
--- a/patches/SLUS-20184_24036809.pnach
+++ b/patches/SLUS-20184_24036809.pnach
@@ -18,8 +18,10 @@ patch=1,EE,00133D54,word,00000000 // remove door animation completed check
 author=illusion
 patch=1,EE,001324D4,word,34640008
 
+[Fix time in computers]
 // Custom sceScfGetLocalTimefromRTC. Adapted from configConvertToLocalTime in PS2SDK.
 author=VelpaChallenger
+description=Fixes time shown when using computers either as Claire or Chris (shows local instead of JP).
 patch=1,EE,001147E8,word,3c19000f	// Save time (sceCdCLOCK) before it's lost. Will be used after we inject. Overwrites printf but meh.
 patch=1,EE,001147EC,word,27391010
 patch=1,EE,001147F0,word,af320000


refractionpcsx2:
You had set the timezone in your BIOS correctly before checking the computers, right? because if it's going off your PS2 time you will need to set your timezone correcly. Just don't want to see a patch being added because somebody missed setting up their BIOS correctly :D

That said some games *do* ignore that and just do japan (looking at you, Marvel Vs Capcom 2)

VelpaChallenger:
I did. In fact, I tried with different timezones and still same result (and with the patch, the right local time was displayed). Plus, I compared with Marvel Vs Capcom 2, and the function they call is identical (except, of course, for the exact RAM addresses/offsets, but the logic is the same to a t).

patches/SLUS-20184_24036809.pnach
```diff
@@ -16,4 +16,308 @@ patch=1,EE,00133D54,word,00000000 // remove door animation completed check
 
 [Skip Startup logos and warning]
 author=illusion
-patch=1,EE,001324D4,word,34640008
\ No newline at end of file
+patch=1,EE,001324D4,word,34640008
+
+// Custom sceScfGetLocalTimefromRTC. Adapted from configConvertToLocalTime in PS2SDK.
+author=VelpaChallenger
```
pgert, line 22: You probably shouldn't have two "author=" entries in the same hack.
And why don't you have this as a separate hack?

[Time fix]
author=VelpaChallenger
description= ...

pgert: I don't quite understand you.
You enable hacks in the "patches" or "cheats" menu,
 depending on were you have placed the pnach-file.
"[Time fix]" won't be enabled by default like "Widescreen 16:9",
 but nighter will "[Skip Startup logos and warning]".

RedDevilus: This is the text we use in Discord

**How do I enable cheats?**
- Settings > Game Properties > Cheats > Enable Cheats
- Tick the boxes in the list for the cheats you want to activate

**Where can I find cheats?**
- <https://gamehacking.org>
- [PCSX2 forums](https://forums.pcsx2.net/Thread-Post-your-PCSX2-cheats-patches-here)

**How do I add cheats?**
Place your pnach file in the `cheats` folder. Your pnach filename should be the serial followed by the CRC (SLUS-12345_1234ABCD.pnach).

**Where is the cheats folder?**
In PCSX2, hit `Tools > Open Data Directory`. A new window will open, and the `cheats` folder will be inside.

**Can I load savestates?**
We suggest **NOT** using savestates with cheats. Cheats can be burned in to or erased from a game by loading savestates.

**Documentation**
- [Pnach 2.0 Documentation](https://forums.pcsx2.net/Thread-Sticky-Important-Patching-Notes-1-7-4546-Pnach-2-0)
- [PS2 Cheat Command Guide](https://github.com/mlafeldt/ps2rd/blob/master/Documentation/code_types.txt)

------------+++---------

A bit annoying is that there is no standardisation on how to name the gametitle= name of game + crc and serial. But that's besides point. 

One link gives detail about version 2 of pnachs which will turn unlabelled patches (everything turned on but you don't know what it does) vs Version2 where you can select a few pieces or everything but category and explanation if you hover as tooltip.

Version 2 will also avoid any CRC collision of other games albeit rare.

PR #601, opened 2025-08-27
Adding patches for God of War II's Bonus Subset (RetroAchievements)

JohnnyTHG:
Added <RAUS-97481_2F123FD8.pnach> as a copy of <SCUS-97481_2F123FD8.pnach> for RetroAchievement's GoWII Bonus Subset.

### DIFF
--- a/patches/RAUS-97481_2F123FD8.pnach
+++ b/patches/RAUS-97481_2F123FD8.pnach
@@ -1,0 +1,16 @@
+gametitle=God of War 2 (RAUS-97481) / [Subset - Bonus]
+
+// This file is a copy of SCUS-97481_2F123FD8.pnach, please keep these two in sync!
+// This game version is a dummy patch for the "Bonus" RetroAchievements subset:
+// https://retroachievements.org/game/28250
+
+[Widescreen 16:9]
+gsaspectratio=16:9
+author=nemesis2000
+patch=1,EE,00234A48,word,46000406
+
+
+[Skip Cutscenes]
+author=Ezedequias
+comment=With Any Action Button
+patch=1,EE,202D8194,byte,01


JohnnyTHG:
The only thing I'm concerned about is the serial. Gollawiz decided to alter the game's serial number from SCUS to RAUS, rather than modifying something irrelevant instead. I hope it won't cause any conflict with PCSX2's database.

MackdannyTwitch:
I remember the Jak games did this initially, but it was such a hassle because without fixes from the database, the RA patched games had visual issues. Eventually the ganes were able to be patche under the proper gameid, but different CRC. I suspect you'll have a better time if that were the case, but we'll have to see what a reviewer says on the matter.

JohnnyTHG:
Exactly! PCSX2 has per-game hardware fixes that are not activated because the patched serial number is unknown. That being said, I've played the game through and it's fortunately no big deal at all. My pull request is just to help the poor player skip minutes long cutscenes every time they want to attempt an OHKO section.

MackdannyTwitch:
> Hi JohnnyTHG, I have a line of communication to Gollawiz and have informed them on how to create a patch for the game which does not change the serial number. When that is completed, it'd likely be cleaner to just take the existing patch file and rename it to drop the CRC hash, i.e. rename <SCUS-97481_2F123FD8.pnach> -> <SCUS-97481.pnach>, which should work with any program that has that serial but isn't necessarily hashed the same. That way, two copies of the same patches don't have to be maintained.



I didn't know this could be done like this. If true, I woukd have done it this way when submitting the patch for the RA versions of the Jak games, and Max Payne 1.

JohnnyTHG:
Thank you for the speedy reply Souzooka. The solution you described couldn't be more perfect at this point. As inexperienced as I am at using git, should I abort the pull request and let you guys work it out? Should I open another pull request renaming the .pnach? Let me know and thank you again.

JohnnyTHG:
A bit late because "reset HEAD~" wasn't really working for me (I'm stoopid :3) so I just git removed the unnecessary copy, git renamed GoWII .pnach file and pushed to my fork. The commit has been successfully added to the pull request. Everything should be fine, but let me know if you feel something is out of place.

Souzooka:
OK, I have to apologize here because while I'm certain someone explained to me that just using the serial in the filename was valid when explaining pnach files, I just tested and it appears not to be valid, so I was mistaken. This might be a useful change to implement into pcsx2 in the future to deal with RA subset files, but does not currently appear to be a feature. Using the retail serial along with the different CRC (and an explanatory comment) is probably still the way to go, as they'll appear next to eachother in the file explorer and could also potentially be deduped if matching only based on serial is implemented.

JohnnyTHG:
That I cannot do as of now because Gollawiz hasn't released a revised patch yet: I don't know what the new CRC will be. The issue will be on hold until then. Thanks for letting me know. Cheers.

PR #614, opened 2025-09-12
Need for Speed - Most Wanted added regions and cheats

Sal89:
Added missing Japanese, Korean and NTSC-U 1.01 versions, plus a new cheat, the "Enable Black Edition" for all non-Black Edition releases apart Korean, that have it enabled by default. As an extra, a "Burger King Challenge" and "Castrol Ford GT" always enabled are added for all standard and Black Edition versions


CookiePLMonster:
Any update on this?

PR #616, opened 2025-09-22
Create SLES-55675_40931C38

vanmatten:
Adds ultrawide patch for PES 2014: Pro Evolution Soccer (Italian & Greek version).
It is the same as SLES-55673_A8A7A42A.

Tested and working.

### DIFF
--- a/patches/SLES-55675_40931C38.pnach
+++ b/patches/SLES-55675_40931C38.pnach
@@ -1,0 +1,15 @@
+gametitle=Pro Evolution Soccer 2014 [E] (SLES-55675)
+
+[Widescreen 16:9]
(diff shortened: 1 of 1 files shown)


pgert:
This hack is not Ultrawide (21:9), but a regular 16:9 Widescreen hack.
Also the G in PAL-G-I stands for German, not Greek.

vanmatten:
I edited the text, I just wrote it wrong. I know it'is 16:9.
Although the G stands for German, I can confirm that the only two languages that can be chosen in-game are Italian and Greek. Maybe is it wrong the database?

If you need I can upload a screenshot later.

PR #620, opened 2025-09-26
Patches: SpongeBob Movie add spanish version + tidy up

CRASHARKI:
Changes:
- Add Widescreen patch to SpongeBob The Movie (Spanish version) to run the game at 16:9 Widescreen Aspect Ratio. It uses the same CRC as another regional version already included, but I tested it just in case.
- Tidy up.


Mrlinkwii (approved):
assumed tested and working

PR #625, opened 2025-10-24
Patches: Add SLUS-21490 and SLES-53753 60FPS

AshMetalRaf:
Added Test Drive Unlimited (NTSC-U) SLUS-21490 60FPS patch (ported from the PAL version).
Retained original PAL-M SLES-53753 50/60FPS patch by PeterDelta & ijandric97.
I've tested both patches and confirmed they're working fine.

Source: https://github.com/PeterDelta/PCSX2/tree/main/patches


MarcoEstevez:
Thanks!!

PR #626, opened 2025-10-25
Update Patches for Silent Hill 2 SLUS-20228 and Silent Hill 3 SLUS-20622

Snake356:
I modified the patch for Silent Hill 2 so that you can disable the noise filter and display a map, which you can activate or deactivate at any time.
I modified the patch for Silent Hill 3 so that you can use a camera like in RE4, and you can also fix the camera angle with the L3 button. This can also be activated and deactivated at any time.

The patches work; I tested them myself and couldn't find any bugs.

### DIFF
--- a/patches/SLUS-20622_2498951B.pnach
+++ b/patches/SLUS-20622_2498951B.pnach
@@ -33,12 +33,12 @@ description=Patches the game to run at 60 FPS
 patch=1,EE,2013058C,extended,00000000
 
 [Camera behind Heather]
-author=RODOLFO NUÑEZ
-description=RE4 Style Camera
+author=Snake356
+description=RE4 Style Camera for SH3
 patch=1,EE,21E7211C,extended,00000001
 
 [Camera Angle fix Press L3]
-author=RODOLFO NUÑEZ
-description=Fix Camera Angle with L3
+author=Snake356
+description=Fix Camera Angle with L3 (RE4 Style Camera)
 patch=1,EE,D03547DC,extended,0000FFFD
 patch=1,EE,01E72114,extended,00000000


pgert:
Silent Hill 2 (v2.01) * NTSC-U * SLUS-20228 * FE06A030

Add author:
author=Snake356 (Rodolfo Nuñez)

Also, I assume that the "black borders fix" is a "No Blackbars" hack.
If so, make it a hack of its own.

---------------------

Silent Hill 3 * NTSC-U * SLUS-20622 * 2498951B

Make author like this instead:
author=Snake356 (Rodolfo Nuñez)
It makes it easier to find the author (you).

Any comment on this hack?

[60 fps]
description=60 fps shortcut: L3=On / R3=Off (performance hit)
patch=1,EE,D035475C,extended,FFFD
patch=1,EE,20130588,extended,34020001
patch=1,EE,D035475C,extended,FFFB
patch=1,EE,20130588,extended,8E220004

Snake356:
> I know you didn't submit the [60 fps] hack, but you can still test it (if you're up for it). I also know you didn't submit any [No Blackbars] hack, but you can still find out if the "black borders fix" is a [No Blackbars] hack, and if so make it a hack of its own (if you're up for it).

Oh, that's what you meant. If I knew more about it, I would definitely do it, no question. But my knowledge is limited, so I usually just complete the games where the codes are missing so that other players can also enjoy the benefits of playing the games. I'm just someone who usually ports and tests the codes from other versions. The codes in the pnach files offered here are already integrated, so they should work normally without any problems, otherwise they wouldn't be entered in the pnach file in advance. That's why I only test the codes that I have added myself.

Snake356:
> > > Silent Hill 2 (v2.01) * NTSC-U * SLUS-20228 * FE06A030
> > > Add author: author=Snake356 (Rodolfo Nuñez)
> > > Also, I assume that the "black borders fix" is a "No Blackbars" hack. If so, make it a hack of its own.
> > > Silent Hill 3 * NTSC-U * SLUS-20622 * 2498951B
> > > Make author like this instead: author=Snake356 (Rodolfo Nuñez) It makes it easier to find the author (you).
> > > Any comment on this hack?
> > > [60 fps] description=60 fps shortcut: L3=On / R3=Off (performance hit) patch=1,EE,D035475C,extended,FFFD patch=1,EE,20130588,extended,34020001 patch=1,EE,D035475C,extended,FFFB patch=1,EE,20130588,extended,8E220004
> > 
> > 
> > I didn't submit any No Blackbars hack, just the No Noise Filter and the map code for SH2. No Noise Filter is missing in SH2 because it's available in other titles from the Silent Hill series with the codes. In SH3 and SH4, you can turn it on and off separately in the game.
> > In Silent Hill 3, the code belongs to the Camera Behind Heather. I don't know where you got the codes from, as I didn't submit the ones you mentioned.
> 
> No Noise Filter in SH2 It's not completely lost, this option is available after beating the game at least once, I had the goal of making a patch with the game's native system (turn off/turn on noise filter) but due to lack of time I have not been able to finish a game again.

I didn't even know you could get that feature. It seems pointless to unlock it only after playing through the game for the first time. I thought the feature didn't even exist in the game, but it's good to know. There must be savegames you could use, then it should be unlocked and you could work on it, so you wouldn't have to play through the whole game again. But my code method is just as good, you can also turn it on and off in the patches if you want. Of course, it would be better if you could turn it on and off directly in the game, that would be the better solution.

Mrlinkwii:
whats the status on this ?

Snake356:
> whats the status on this ?

So far, so good, the codes work without any problems.

Mrlinkwii (approved):
assumed tested

PR #627, opened 2025-10-28
Added WS fixes for other versions

YukiXXL:
Added WS fixes for these versions of the "Syphon Filter: The Omega Strain" game:
Public Beta 1.0
Beta Trial Code
November Demo
Jampack Demo
Review Prototype


F0bes (changes requested):
Please remove the SCUS-97397 prototype.

RedPanda4552:
We had a look at these, checks out that it was a public release.

GovanifY:
Since it's a public beta it could be legally acquired, so I don't really see any reason to remove it personally.

PR #630, opened 2025-11-05
Removed comments from Ratchet & Clank widescreen patch

thatonedood:
SCUS-97199_CE4933D0.pnach
NTSC-U

The pnach that was in place had comments in all of the code, so I reuploaded the pnach without comments so that PCSX2 could read the file.


thatonedood:
Also this is my first time doing this. (like contributing to anything on git). If I did something wrong, lemme know.

thatonedood:
My mistake, didn't see that. Thanks. Will learn from this.

PR #632, opened 2025-11-11
NASCAR Dirt to Daytona patches

CookiePLMonster:
Adds a few patches for **NASCAR: Dirt to Daytona** (SLUS-20441):

* 60 FPS. Flawless, as both NASCAR Heat 2002 and Test Drive: Eve of Destruction are 60 FPS games, but may need a mild overclock.
* Shoulders control mapping. GameCube/Xbox-style controls mapping with throttle/brake on triggers.
* Analog camera control
* Extended valid birth year range

### DIFF
--- a/patches/SLUS-20441_2EA87CC5.pnach
+++ b/patches/SLUS-20441_2EA87CC5.pnach
@@ -1,0 +1,61 @@
+gametitle=NASCAR - Dirt to Daytona [SLUS-20441] (U)
+
+[60 FPS]
+description=Use EE 180% overclock to minimize drops during big crashes
+author=Silent
+
+patch=0,EE,20155DB8,extended,00000000
+
+[Shoulders control mapping]
+comment=Replace Alternate2 controls mapping with Shoulders controls mapping from NASCAR Heat
+author=Silent
+
+// Patch controls
+patch=0,EE,10251108,extended,7D40 // throttle = Joy DY-
+patch=0,EE,20251110,extended,264600F8 // dig_throttle = Button 8
+patch=0,EE,10251134,extended,7D48 // braking = Joy DX-
+patch=0,EE,1025114C,extended,7D10 // dig_braking = Button 7
+patch=0,EE,1025117C,extended,7C38 // downshift = Button 1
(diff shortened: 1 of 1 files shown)


Gabominated:
Yesterday I just posted 60fps patch for this game at the forum, my address 00155db4

CookiePLMonster:
> Yesterday I just posted 60fps patch for this game at the forum, my address 00155db4

Wouldn't removing the branch be preferable to removing its condition?

CookiePLMonster:
> but I am interested in discussing whether your code is better than the one I mentioned before.

The intent is to remove the check that skips physics sim on odd frames, so I feel like skipping the branch altogether is a better solution than modifying the condition.

Gabominated:
> > OK, but changing the address value 00155db4 to 38420001, I personally tested this value and it works without problems, so we could say that both addresses achieve the same result of releasing the frames without causing collateral damage.
> 
> Well, with this you're changing the check into `(simTickCount ^ 1) == 0` which will only ever be true if `simTickCount` is `1`. However, the public version of your patch turns it into `simTickCount == 0` by nopping the `andi` instruction.
> 
> Both are technically fine, but if you want to remove the condition entirely (which is the goal here), why not just remove it instead of coming up with different ways of making it an always-false condition?

I suppose it will depend on the developer's work style and preferences. I have a couple of degrees here that, to date, I haven't been able to achieve anything with the knowledge I have, but you know more than me and you could succeed quickly. 
- Finny the Fish & The Seven Waters

Don't forget the NASCAR widescreen patch; I've tested it extensively and it works well in 16:9.

CookiePLMonster:
> OMG, I missed that, forget it, maybe for ultrawide but i really don't know

I'd be in favour of submitting ultra-widescreen patches, but they can also be done even better than the current approach - UI margins and even the display string in the options menu can be modified, which these patches are currently not doing.

The string is huge nitpick I'd not expect to be changed, but UI margins could be easily adjusted to make the HUD span the entire screen. The game already gives "negative margins" to the UI for 16:9.

GovanifY (approved):

PR #633, opened 2025-11-11
Add patches for Shin Megami Tensei: Digital Devil Saga 1, 2, and Noct…

Josh1billion:
…urne


patches/D7273511.pnach
```diff
@@ -0,0 +1,35 @@
...
+patch=1,EE,D03F9902,extended,0000FEFE
+patch=1,EE,2011CC88,extended,A4831364
+
+
+//enable both of these as a pair by removing the //
+
+// Atma Gain X2
+patch=1,EE,201A7B98,extended,00118840
+
+// Random Battles Occurrence Halved
```
Mrlinkwii: why was this added ? this has nothing to with the widescreen patch

PR #634, opened 2025-11-12
Test Drive Unlimited: Trigger controls mapping

CookiePLMonster:
A cheat code re-enabling a previously unused control mapping in Test Drive Unlimited, and swapping L1/R1 buttons for L2/R2 triggers. With this change, this mapping matches “modern” controls of the X360 and PC versions of Test Drive Unlimited.

Also added support for `SLES-54466`, as it has the same executable as `SLES-53753`, just a different serial.


pgert:
Have you invented a new word?
"mappping" -> "mapping".

CookiePLMonster:
Funny enough, this is the not the first time I made this typo. Will squash commits while merging so this typo will vanish for good.

PR #643, opened 2025-12-15
Add to SLUS-20973_4028A55F.pnach

UnstableZero:
Add a patch for 16:9 Widescreen to fix the positions of characters, and their pedestals in the inventory.

No AI was used in the making of this commit.

### DIFF
--- a/patches/SLUS-20973_4028A55F.pnach
+++ b/patches/SLUS-20973_4028A55F.pnach
@@ -30,17 +30,6 @@ patch=1,EE,101C0AEC,extended,0000C268 // 0000C22D - Player2 X
 patch=1,EE,101C1040,extended,0000C423 // 0000C3F7 - Pedestal2 X
 
 [No-Interlacing]
-gsinterlacemode=1
-author=Agrippa
-description=EE overclock is needed to avoid infrequent frame rate drops.
-patch=1,EE,2044FD40,extended,00000000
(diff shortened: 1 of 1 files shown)


UnstableZero (commented):
New pedestal positions in widescreen look fine even on larger displays. This could probably be adapted to other aspect ratios easily. Added a No-Interlacing patch I have been using that doesn't force the game into Progressive Mode, and therefore 30FPS mode; old patch skips a WaitSema to achieve 60FPS again.

UnstableZero:
Not opposed to the attribution change. I read the forum post you linked.

This rabbit hole goes really deep, and it appears Agrippa and I have different ideas. As I understand it: FullframeMode 1 and 2 are progressive mode. One flag changes the displayed resolution, and the other triggers the 30FPS mode to allow the engine to properly render each frame in time. Agrippa is using both, and a hack to force 60FPS again(and patching AIstepcount). My patch uses one of the same flags to force the correct resolution, but doesn't force the engine into 30FPS mode and adding extra processing to the rendered image, and hacking speed back. 

While I use a flag to set resolution properly, I am not loading the byte the engine uses to track which offset it needs to render(it flips 0/1) every frame. From a performance stand-point, mine is superior, areas in both games known to be problematic, are in the vast majority of cases, more performant, even without the 133% overclock Agrippa requires. (Caves of K'Thuun in the first game is a great example testing ground.)

PR #645, opened 2025-12-19
Update and at to SLUS-20565_90E66BC5.pnach "Champions of Norrath"

UnstableZero:
Added revised widescreen patch from the forum. This includes 16:9 Inventory Fix as well, this fix can be ported to the old widescreen if there is demand. Added new No-Interlacing patch that does not require EE overclock.

### DIFF
--- a/patches/SLUS-20565_90E66BC5.pnach
+++ b/patches/SLUS-20565_90E66BC5.pnach
@@ -33,17 +33,6 @@ patch=1,EE,20480568,extended,C296C8B4 // C22DC8B4 - Player2 X
 patch=1,EE,101AD7A8,extended,0000C452 // 0000C3F7 - Pedstal2 X
 
 [No-Interlacing]
-gsinterlacemode=1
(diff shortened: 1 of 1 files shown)


PeterDelta:
We are the looking for the simplicity and pragmatism, keeping the end user in mind so they don't have to experiment or try to figure out which patch to use. That's why there's only one patch for widescreen, FPS, etc. My suggestion is that if the interlacing patch is bug-free and improves upon the existing one, replace it. If it complements the existing patch, add it, as you wisely did with the widescreen patch.

UnstableZero:
Given the nature of the No-Interlacing patch currently in use being both No-Interlacing and a Restore to 60FPS hack, I will be replacing it with my own. Some users may notice a difference in on-screen text (sprites) because the game is no longer forced into progressive mode (a feature that is not used in these specific Snowblind Engine games). The case is the same for https://github.com/PCSX2/pcsx2_patches/pull/643

PR #646, opened 2025-12-19
Add widescreen and 60 FPS patches for Sonic Unleashed

thatonedood:
Added new widescreen (16:9) and FMV aspect ratio fixes, as well as a 60 FPS patch for Sonic Unleashed (SLUS-21846). Includes author credits and descriptions for each patch.

I tested these new codes and they work. No crashing.
Codes work during night stages and boss fights. 

Only caveat is that during Dark Gaia Phoenix boss level, the 60 FPS patch has some strange behavior, but it is still playable. There are model animation pauses and a few changes in timing, but I was still able to defeat the boss without any other changes in behavior,
Note: this is only the 60 FPS patch. When I tested the widescreen patch alone without 60FPS, the Dark Gaia Phoenix boss went smoothly with no issues.

### DIFF
--- a/patches/SLUS-21846_FB236A46.pnach
+++ b/patches/SLUS-21846_FB236A46.pnach
@@ -1,3 +1,23 @@
+gametitle=Sonic Unleashed NTSC-U SLUS-21846 FB236A46
+
+[Widescreen 16:9]
+gsaspectratio=16:9
+author=Gabominated
+description=Widescreen hack
+patch=1,EE,00870048,word,3FD21DAF //3F9D9643
+
+[FMVs aspect ratio]
+author=Monsterjamp & Gabominated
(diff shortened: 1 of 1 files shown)


Gabominated:
Thanks for the full report, I only tested it up to the first boss and everything seemed to be going well.

thatonedood:
Oh cool, it’s the creator of the patch. I’m new to all this, so that’s pretty neat.

I should add that I didn’t test the entire game. I searched online for areas where people reported issues, and the only problems I found mentioned were in night stages, boss fights, and specifically the Dark Gaia Phoenix. Because of that, I only selectively tested a few boss fights, a couple of night stages, and the Phoenix. I just don’t have the time or dedication to test everything 100%.

If these patches are implemented, I’d be much more motivated to play through the entire game, since I could do so with RetroAchievements hardcore mode enabled. That’s my main motivation.

PR #650, opened 2025-12-29
SLUS-20565 and SLUS-20973 cleanup.

UnstableZero:
Fix author tag.
Update patch comments to match functionality.
Add missing gsinterlace setting.
Descriptions match now.
Remove the 30FPS flag, as the engine may still
want to drop to 30FPS in certain situations.
Widescreen patches are heavy in this game engine, and may introduce lag, especially when coupled with Frustum Fixing patches. It may be the case now that the new No-Interlacing patch doesn't require an EE overclock, users will see lag when using Widescreen, and or the game engine utilizing its inbuilt 30fps mode to keep gameplay smooth.


PeterDelta (approved):

UnstableZero:
Closing this PR to fix commit history and add a Y-offset fix, keen eyed players have noticed a single pixel alignment issue, that gets multiplied (and harder to see) when upscaling.

PR #671, opened 2026-02-17
Patches: Add Monsters, Inc. Scare Island Widescreen to all regions

CRASHARKI:
Added the Widescreen patch to all other regions + some tidy up.
As you can tell by the CRCs, the codes are the same for all regions, still, all of them were tested just in case.


patches/SCES-50596_F5B07CC0.pnach
```diff
@@ -0,0 +1,17 @@
+gametitle=Disney/Pixar Monsters, Inc. - Skraemmeoen (PAL-D) (SCES-50596)
+
```
pgert: "Skraemmeoen" -> "Skræmmeøen".
http://redump.org/disc/76730/

CRASHARKI: Well, looks like a simple "é" breaks the system but these characters get a pass. Thanks for pointing out that the other files already included these characters, I just copied the names and didn't pay much attention.

Mrlinkwii (approved):

PR #673, opened 2026-02-19
Sly Trilogy NTSC-U No-Interlacing patch name fixes

Windsurf7:
It seems that the No Interlacing & No Motion Blur patch merging was done wrongly for Sly 2, as opposed to what can be observed in Sly 1 and 3 to see how it was supposed to be done.

This also causes the patch not to get detected by the "Apply No-Interlacing Patches" option.

Please inspect the required fixes as proposed.

Also added missing author labels.


Windsurf7:
The other mentioned fixes have been merged so this is no longer necessary. Closing.

PR #682, opened 2026-03-08
Patches: Add remaining patches to The Warriors

CRASHARKI:
All patches from the PAL version were added to the NTSC-U version, along with a new patch I created to enable widescreen by default in both versions. The NTSC-U viewport width fixes turned out to be unnecessary, as the difference is barely noticeable, and the 4:3 mode itself seems slightly narrower, while 16:9 matches the PSP version's width. In addition, those fixes narrowed and misaligned the HUD, which was actually noticeable. Fortunately, the memory addresses were identical in both versions, although both were tested to confirm compatibility.

### DIFF
--- a/patches/SLES-53443_FA3C1346.pnach
+++ b/patches/SLES-53443_FA3C1346.pnach
@@ -1,4 +1,13 @@
-gametitle=Warriors, The (PAL-M) SLES-53443 FA3C1346
+gametitle=The Warriors (PAL-M) (SLES-53443)
+
+[Widescreen 16:9]
+gsaspectratio=16:9
+author=CRASHARKI
+description=Run the game at 16:9 Widescreen Aspect Ratio from the start.
+
+//Widescreen hack 16:9
+//Enable ingame widescreen
+patch=1,EE,007AA05C,byte,01 //00
 
 [50/60 FPS]
 author=PeterDelta
--- a/patches/SLES-53443_FA3C1346.pnach
+++ b/patches/SLES-53443_FA3C1346.pnach
@@ -3,16 +3,14 @@ gametitle=The Warriors (PAL-M) (SLES-53443)
 [Widescreen 16:9]
 gsaspectratio=16:9
 author=CRASHARKI
-description=Run the game at 16:9 Widescreen Aspect Ratio from the start.
-
-//Widescreen hack 16:9
-//Enable ingame widescreen
+description=Enable native widescreen.
 patch=1,EE,007AA05C,byte,01 //00
 
-[50/60 FPS]
-author=PeterDelta
-description=Might need EE Overclock at 130%.
-patch=1,EE,00596D70,byte,01
+// Apparently doesn't achieve real 50/60 FPS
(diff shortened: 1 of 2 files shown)


CRASHARKI:
> Neither of the 50/60 fps patches works, there's only a false reading because the frame limit is unlocked, but the game remains at 25/30 fps. My suggestion is to remove both patches to avoid confusion.

I've been playing with the patch enabled and it feels smoother to me than without it; plus, it's harmless, as I haven't had any issues.
I think it'd be appropriate to include the patch creator, @PeterDelta, in the conversation.

Gabominated:
> > Neither of the 50/60 fps patches works, there's only a false reading because the frame limit is unlocked, but the game remains at 25/30 fps. My suggestion is to remove both patches to avoid confusion.
> 
> I've been playing with the patch enabled and it feels smoother to me than without it; plus, it's harmless, as I haven't had any issues.
> I think it'd be appropriate to include the patch creator, @PeterDelta, in the conversation.

first without the patch and then with the patch activated, upload it here and let's see how much the difference is. I don't see any

CRASHARKI:
I played the game from beginning to end and had no issues at all. The experience was definitely smoother than without the patch.

Gabominated:
> I played the game from beginning to end and had no issues at all. The experience was definitely smoother than without the patch.

With all my experience, I can say that there is no difference, everything is the same.

CRASHARKI:
> With all my experience, I can say that there is no difference, everything is the same.

The purpose of this patch is to increase the frame rate from 30 FPS to 60 FPS. That behavior can be measured and verified objectively. If you personally don't perceive a difference, that's perfectly fine, but the patch is functioning as intended.

Gabominated:
> > With all my experience, I can say that there is no difference, everything is the same.
> 
> The purpose of this patch is to increase the frame rate from 30 FPS to 60 FPS. That behavior can be measured and verified objectively. If you personally don't perceive a difference, that's perfectly fine, but the patch is functioning as intended.

The purpose of a good patch is to achieve a true 60 fps, and this patch doesn't. It only achieves a false reading on the emulator's FPS meter. To support what I've been saying, the author himself, @PeterDelta , came to me to verify why the patch wasn't achieving 60 fps. I've spent thousands of hours trying to create a true 60 fps patch, but I haven't succeeded yet. The closest I've come is an unplayable patch that you can find in my GitHub collection. This patch does achieve 60 fps, but with many bugs. The game is very difficult.

CRASHARKI:
> The purpose of a good patch is to achieve a true 60 fps, and this patch doesn't. It only achieves a false reading on the emulator's FPS meter. To support what I've been saying, the author himself, @PeterDelta , came to me to verify why the patch wasn't achieving 60 fps. I've spent thousands of hours trying to create a true 60 fps patch, but I haven't succeeded yet. The closest I've come is an unplayable patch that you can find in my GitHub collection. This patch does achieve 60 fps, but with many bugs. The game is very difficult.

Okay, since you are working on a better patch I'll remove this one.

PR #696, opened 2026-03-30
Separate "No Interlacing" and "No Motion Blur" patches for the Sly Cooper games + demos

SlyCooperReloadCoded:
Several years ago, all games and demos in the Sly Cooper series had two patches each, not including the widescreen patches:

1. No Interlacing, a patch that disabled the screen shaking caused by turning off deinterlacing.

2. No Motion Blur, a patch that disabled a game-specific blur-like function.

At some point in time, these two types of patches got confused, and most if not all of them got deleted, mislabelled, duplicated, and so on. PR #668 addressed this issue, re-adding all the patches and correcting authors. However, it chose to intentionally combine the two types of patches because they both relate to image quality, however this is incorrect in practice and people I showed this to were enraged. This PR separates them into the separate patches they should be, restoring them as they were before the initial screw-up several years ago.

I also fixed a few small typos here and there that I noticed. Let me know if I did anything wrong, it's been a while since I've made a PR.

### DIFF
--- a/patches/PAPX-90231_9C29F787.pnach
+++ b/patches/PAPX-90231_9C29F787.pnach
@@ -1,6 +1,6 @@
 gametitle=怪盗 スライ・クーパー [体験版] (PAPX-90231)
 
-[No Interlacing]
+[No-Interlacing]
 gsinterlacemode=1
 author=asasega
 description=No Interlacing
--- a/patches/PAPX-90231_9C29F787.pnach
+++ b/patches/PAPX-90231_9C29F787.pnach
@@ -1,12 +1,16 @@
 gametitle=怪盗 スライ・クーパー [体験版] (PAPX-90231)
 
-[No-Interlacing]
+[No Interlacing]
 gsinterlacemode=1
-author=asasega & Meos
-description=No Interlacing & No Motion Blur
+author=asasega
+description=No Interlacing
 
 // No Interlacing by asasega
 patch=1,EE,20208ED4,extended,24060000
+
+[No Motion Blur]
+author=Meos
+description=No Motion Blur
 
 // No Motion Blur by Meos
 patch=1,EE,201626CC,extended,24060000
--- a/patches/SCED-51452_F3FD8A14.pnach
+++ b/patches/SCED-51452_F3FD8A14.pnach
@@ -1,6 +1,6 @@
 gametitle=Sly Raccoon [Demo] (SCED-51452)
 
-[No Interlacing]
+[No-Interlacing]
 gsinterlacemode=1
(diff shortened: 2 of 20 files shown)


Mrlinkwii:
i believe this fixes #669

SlyCooperReloadCoded:
#668 actually addressed #669 once it completed, but it opted to merge the patches despite protest, so I did it myself. All the patches are now there, this just separates them.

pgert:
Why "[No-Interlacing]" -> "[No Interlacing]"?
Will "Apply No-Interlacing Patches" read this?

SlyCooperReloadCoded:
Oh? I thought that was a typo, I didn't know that option looked for a specific spelling. Does it? My bad if so.

pgert:
I'm pretty sure that it is with the specific spelling.
Revert "[No Interlacing]" back to "[No-Interlacing]".

SlyCooperReloadCoded:
Okay, I believe it's now fixed.

CRASHARKI:
> people I showed this to were enraged

My PR stood there for more than a month before it got accepted and everybody was fine with it since nobody bothered to comment. Plus, having the no interlacing + no motion blur together is a common practice I've seen in other patches from the repo.

> I also fixed a few small typos here and there that I noticed. Let me know if I did anything wrong, it's been a while since I've made a PR.

You call removing my name from the credits of patches I ported "fixing typos"? That is unnacceptable.
And you also removed lines of code which I guess was a mistake too. This is a mess.

RedPanda4552:
@CRASHARKI It appears your authorship was moved to the relevant patch, not removed? And if lines were mistakenly croped out, there are more polite ways to call out that there are errors. Please stay civil and don't resort to trying to start drama.

SlyCooperReloadCoded:
The typos were mainly word being used in place of extended in a few places, a few extra lines here and there that had since been discovered to be obsolete, and the dash in No-Interlacing before I found out it was functional, which I've since reverted. As for the credit, asasega had already found all of the No-Interlacing and No Motion Blur patches for all the builds and submitted them before the initial mix-up. I didn't intend on removing credit, rather I just put them back in using the format they were in when they were intially submitted by the original contributor. I actually still had them on hand as I choose to use them in PNACH form rather than rely on Patches. No harm was intended.

SlyCooperReloadCoded:
I gave the patches.zip from this PR a try, and I can confirm that they all appear in the Patches menu correctly and work as they should. Is there anything else I need to do before this can be merged?

Capture

Capture2

Capture3

Mrlinkwii (approved):

PR #698, opened 2026-04-11
Add NTSC mode patch to Rayman M (v2.00)

RibShark:
While this game has an NTSC version (Rayman Arena), there are a significant number of changes in that version which make it (IMO) inferior to the PAL release. Such changes include the removal of easter eggs, obstacles and shortcuts in race mode, as well as the addition of a freeze bullet mechanic that can be annoying. In battle mode, the Flame Tongue item was removed on the NTSC version.

This patch makes the PAL release run in NTSC mode at 640x448 internal resolution. I have tested this and the game still runs at the correct speed, with no graphical errors (there was actually a leftover function to switch to this mode, I basically recreated that). Not sure if any other patches like this currently exist, but it seems a worthwhile addition.

### DIFF
--- a/patches/SLES-50457_DDA2FA6A.pnach
+++ b/patches/SLES-50457_DDA2FA6A.pnach
@@ -1,7 +1,19 @@
-gametitle=Rayman M (PAL-M 2.0) (SLES-50457)
+gametitle=Rayman M (PAL 2.0) (SLES-50457)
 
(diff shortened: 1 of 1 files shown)


pgert:
"author=RibShark" ?

RibShark:
> "author=RibShark" ?

completely forgot to add that, added it and a description

PR #701, opened 2026-04-13
Add NTSC Mode patch to Theme Park World

Souzooka:
Adds a patch to Theme Park World (EU) which allows the game to run at 30fps.

NOTE: The game is susceptible to framerate differences. While animations seem to run at the correct speed even if the framerate is uncapped (60fps), other mechanics such as the camera speed and how fast time progresses are faster at higher framerates. I am not sure if these mechanics were touched for PAL (25fps), though I assume that 30fps is the intended speed in this case (as it is in Theme Park Roller Coaster).

### DIFF
--- a/patches/SLES-50032_9CBA74B7.pnach
+++ b/patches/SLES-50032_9CBA74B7.pnach
@@ -1,0 +1,8 @@
+gametitle=Theme Park World (SLES-50032)
+
+[NTSC Mode]
+author=Gabominated
+description=Runs game in NTSC Mode (60hz).
+
+// Change vertical height argument to 448. The game handles the rest.
+patch=0,EE,202314FC,extended,240501C0 // addiu a1,zero,0x1C0


Gabominated:
[https://github.com/Gabominated/PCSX2/blob/main/PCSX2%20Patches/SLES-50032_9CBA74B7.pnach](https://github.com/Gabominated/PCSX2/blob/main/PCSX2%20Patches/SLES-50032_9CBA74B7.pnachurl)

Souzooka:
I'm not quite sure what solution would be amicable here; I opened the game and found and implemented the patch myself.

If you'd like to make a pull request with your patch file that's fine, I don't care much either way as long as the patch is available to use alongside RetroAchievements.

Gabominated:
I usually have a list in my repository and also a list in the forum to prevent more than one developer from unnecessarily working on the same patch.

Gabominated:
And the reason I don't make pull requests is because I don't have the opportunity to test all my patches.

Gabominated:
I think the best way to avoid doing this work that's already been done is for you to come back to the forum. I remember seeing your name on the forum talking to asasega about patches in the 60fps thread. I personally post a lot of new patches for people to test and report any problems. Again, I have a pretty up-to-date list, and it only takes a couple of clicks to check for a patch. You're a very good developer, no doubt about it. I confess I've learned a lot from your YouTube videos, they're great tutorials.

Souzooka:
I went ahead and set you as the author; the potential for 60fps did intrigue me (I noticed that variable as well), though a patch certainly wouldn't be suitable to release as is. It might not take too much to polish it up, though. 

NTSC Mode seems fine to me, though I did notice that the game had pretty bad cropping on both NTSC and PAL modes. I'll have to see if the US version is like this too.

Gabominated:
Yeah, no doubt about it, The FPS patch needs fixing and testing, but I usually post it so someone can tell me how playable the patch is; in fact, that patch was requested of me here [https://github.com/Gabominated/PCSX2/issues/27](https://github.com/Gabominated/PCSX2/issues/27)

Souzooka:
@Gabominated If you want to test it with your patch, changing 0x1C4AD8 so that 0x2000 is written instead of 0x4000 seems to cause all other systems (how fast days increment, camera movements, some motions) to all run at the correct speed. This value is also 0x4000 on the US version, so 0x4000/30fps can probably be considered the developer intent.

Presumably they're doing some sort of timing thing in this function, saving a value to this variable and if it's greater than 0x4000, just saving 0x4000 to the variable. For whatever reason though, that value seems to be always 0x138800 for me in the PAL version, which is way off the mark and pretty curious.

The game seems rather CPU-intensive at 60fps though, unfortunately.

Gabominated:
> @Gabominated If you want to test it with your patch, changing 0x1C4AD8 so that 0x2000 is written instead of 0x4000 seems to cause all other systems (how fast days increment, camera movements, some motions) to all run at the correct speed. This value is also 0x4000 on the US version, so 0x4000/30fps can probably be considered the developer intent.
> 
> Presumably they're doing some sort of timing thing in this function, saving a value to this variable and if it's greater than 0x4000, just saving 0x4000 to the variable. For whatever reason though, that value seems to be always 0x138800 for me in the PAL version, which is way off the mark and pretty curious.
> 
> The game seems rather CPU-intensive at 60fps though, unfortunately.

Interesting, this looks better. Unfortunately, I don't have much time to do testing, but go ahead if you want to finish this patch.

F0bes:
Is this ready to be merged, or is there more work to be done?

Gabominated:
> Is this ready to be merged, or is there more work to be done?

Yeap, that´s it for now

Issue #704, opened 2026-04-15
Jimmy Neutron: Attack of the Twonkies (SLUS-20887_058849D1) - Skippable Cutscenes patch causes softlocks


SlyCooperReloadCoded:
Now that this game can be reliably played using a patch to enable its internal HostFS mode, avoiding the constant freezes and lockups (context: https://github.com/PCSX2/pcsx2/issues/10957#issuecomment-3181413598), the patches can be tested more thoroughly. It appears that the "Skippable Cutscenes" patch in the patches database not only automatically skips some cutscenes without any button presses, but it also causes softlocks whenever you make an Invention or Super Invention in the game. Example softlock:

https://www.youtube.com/watch?v=u1JgsBcUkbU

I can confirm that the "Skippable Cutscenes" patch is the sole cause of this bug. Its author will need to tweak it, but it should be removed in the meantime. @Souzooka your assistance is required!

Souzooka:
This is really a game that keeps on giving, and doesn't seem to want to behave does it. It's been a while since I made the patch, but if I recall correctly any patched code should only be running during cutscenes. Is there any understanding why the softlock occurs?

Souzooka:
I'm also admittedly not too familiar with the game but if creating an invention is supposed to start a cutscene, I can see how that might cause the issue (if the cutscene actually just gets skipped immediately).

If this is the case, editing the patch to only allow skipping after the cutscene has been playing for a second or so might be an easy enough fix.

SlyCooperReloadCoded:
Skipping after a second or two is also how most games that have skippable cutscenes seem to do it.

Souzooka:
I'll take a look today and see if I get the same issue. If this is what is causing the issue, setting the length to a whatever is smaller, the length or two seconds, instead of zero should fix it hopefully.

(Or well, we'll have to make sure this doesn't happen on the first frame too, I'll poke it).

Souzooka:
OK, I've reproduced the issue. Holding X when going into the cutscene does softlock the game; this should be easy to prepare a fix for.

Souzooka:
This issue should be corrected in the above PR.

Mrlinkwii:
fixed with #705

PR #705, opened 2026-04-15
Fix potential softlock with Skippable Cutscenes patch in Jimmy Neutron: Attack of the Twonkies (US)

Souzooka:
Holding X/Start when inventing could cause the player to get stuck in the Inventerface, so this forces the cutscene to play for a bit before allowing skipping.

Addresses https://github.com/PCSX2/pcsx2_patches/issues/704


Souzooka:
Video showing me holding X to skip the cutscene, then reloading and letting the cutscene play in full:

https://github.com/user-attachments/assets/b1ff2651-fbdb-4200-9314-ab339569e444

SlyCooperReloadCoded:
This appears to work fine for me!

Mrlinkwii (approved):

PR #709, opened 2026-04-21
Ratchet & Clank 3 (SCES-52456): add widrescreen patch for remaining locations

ricnava00:
Added the remaining locations, as the patch is level-specific

The Vid-comic patches may not be as necessary, and some vertical levels like the 5th one will show some unintended graphics at the sides (this already happens for the NTSC patch)


Heitof:
Can't you use dynamic patches instead?

ricnava00:
I kept it the same as before to reduce the possibility of breaking something, but this should work.
The only downside is that if the patch is enabled while the game is running, it will crash, since the dynamic patch gets applied before the static code is added and so it will jump to nops.
If someone knows how to correct this please do

patches/SCES-52456_17125698.pnach
```diff
@@ -4,6 +4,11 @@ gametitle=Ratchet & Clank 3 (PAL-M5) (SCES-52456)
 gsaspectratio=16:9
 author=ElHecht
 
+dpatch=0,3,3,0,c46000b0,4,46010002,14,1460000a,0,0c03ff80,4,00000000,14,1060000a // jump to code-inject 200ffe00
+dpatch=0,4,4,0,00000000,4,00000000,10,00000000,30,10400008,0,3c013f40,4,4481f000,10,461e1082,30,14400008 // hor fov hud and menu
+
+dpatch=0,3,3,0,0803ff80,4,00000000,14,1060000a,0,0c03ff80,4,00000000,14,1060000a // fix savestates with old patch
```
ricnava00, line 10: I've added this line to fix savestates that used the old patch, I don't know what the policy is in these cases, it can safely be removed otherwise

chaoticgd:
>  The only downside is that if the patch is enabled while the game is running, it will crash, since the dynamic patch gets applied before the static code is added and so it will jump to nops.

Yes that's the problem with mixing dynamic patches and regular patches. You could add some logic to make sure that your code cave exists, but that's quite hacky, so I think it would just be best to avoid it if you can.

Update: I've thought of a way to fix this behaviour.

F0bes:
Could you add that note to the patch description? Something like "Don't apply while the game is running"

PR #711, opened 2026-04-22
Killzone (SCUS-97402) 480p Progressive Scan Patch

mthsk:
The wiki says this game supports Progressive Scan but I couldn't find information on how to enable it as the game lacks an in-game setting and the Cross + Triangle button combination didn't work for me.

Regardless, here is a progressive scan patch.


Mrlinkwii (approved):
assumed tested

PR #715, opened 2026-05-09
Remove GTA Stories No-Interlacing patches

CRASHARKI:
The original submission had several issues:
- The [No-Interlacing] tags were wrongly written as [No Interlacing] making them not work with the global enabler.
- The values were changed from 0 to 1.
- GTA Vice City Stories wasn't even using the right patch, it was using GTA Vice City's patch.

So I fixed these issues and also ported the patches to PAL versions.

### DIFF
--- a/patches/SLES-54135_D693D4CF.pnach
+++ b/patches/SLES-54135_D693D4CF.pnach
@@ -65,3 +65,9 @@ patch=1,EE,203D8DAA,byte,1
 author=Snake356
 description=Patches the game to run at 50 FPS (Might need 180% EE Overclock to be stable).
 patch=1,EE,202D9384,word,00000000 //14820017
+
+[No-Interlacing]
+author=Mensa; ported by CRASHARKI
+description=Attempts to disable interlaced offset rendering.
+gsinterlacemode=1
+patch=1,EE,2035F394,word,00000000 //64420008
--- a/patches/SLES-54135_D693D4CF.pnach
+++ b/patches/SLES-54135_D693D4CF.pnach
@@ -66,8 +66,9 @@ author=Snake356
 description=Patches the game to run at 50 FPS (Might need 180% EE Overclock to be stable).
 patch=1,EE,202D9384,word,00000000 //14820017
 
-[No-Interlacing]
-author=Mensa; ported by CRASHARKI
-description=Attempts to disable interlaced offset rendering.
-gsinterlacemode=1
-patch=1,EE,2035F394,word,00000000 //64420008
+//The following patch was discarded as it was deemed unnecessary:
+//[No-Interlacing]
+//author=Mensa; ported by CRASHARKI
+//description=Attempts to disable interlaced offset rendering.
+//gsinterlacemode=1
+//patch=1,EE,2035F394,word,00000000 //64420008
--- a/patches/SLES-54622_B3AD1EA4.pnach
+++ b/patches/SLES-54622_B3AD1EA4.pnach
@@ -24,3 +24,9 @@ patch=1,EE,21FB5830,byte,1
 author=Snake356
 description=Patches the game to run at 50 FPS (Might need 180% EE Overclock to be stable).
 patch=1,EE,203704F4,word,00000000 //14820019
+
+[No-Interlacing]
+author=Mensa; ported by CRASHARKI
+description=Attempts to disable interlaced offset rendering.
+gsinterlacemode=1
+patch=1,EE,204481E4,word,00000000 //64420008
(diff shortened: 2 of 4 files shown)


fox140cv:
PCSX2 fixed the interlacing problem in most games years ago, and these patches can be removed since the image no longer jitters. They often cause jagged edges.

CRASHARKI:
@fox140cv Are you saying that No-Interlacing patches are not necessary in general or that these ones specifically aren't necessary? Because a lot of games still need these patches to actually look right.

fox140cv:
Yes, I'm referring to those specific ones, and generally speaking, because PCSX2 already does a good job. You can compare those interlacing patches with them enabled and then disabled with PCSX2's automatic interlacing; you'll see that either there's no difference, or it looks better without those patches. In either case, there's no improvement.

CRASHARKI:
Okay, so what do you suggest we do here? Should I remove them?

fox140cv:
If there's no improvement, there's no point in having a patch that supposedly does something the emulator already does very well. A few years ago, I removed several patches of this type that disabled that function. There are more examples of what I mean: https://github.com/PCSX2/pcsx2_patches/pull/273
And this happens in most games, there are few that need an interlacing patch today.

CRASHARKI:
Then I'll change the PR to make it a removal of the patches. And thank you for the link to the examples, it was very interesting to check.

I'll leave the patches commented as a reminder.

Immersion95:
> Yes, I'm referring to those specific ones, and generally speaking, because PCSX2 already does a good job. You can compare those interlacing patches with them enabled and then disabled with PCSX2's automatic interlacing; you'll see that either there's no difference, or it looks better without those patches. In either case, there's no improvement.

That's not true for all games. Many games that have rapid flicker will display lines with the Auto/Adaptive deinterlacing, the No-interlacing patches are necessary for those.

Example : Cyclops' ray that is following his eyes in MVC2 or the red lifebar in DOA2. The only thing is that the No-interlacing patch has to be correctly done. It's usually easily implemented in most games but some (Bloody Roar 3) will have many interlacing offset mechanisms.

fox140cv:
> > Yes, I'm referring to those specific ones, and generally speaking, because PCSX2 already does a good job. You can compare those interlacing patches with them enabled and then disabled with PCSX2's automatic interlacing; you'll see that either there's no difference, or it looks better without those patches. In either case, there's no improvement.
> 
> That's not true for all games. Many games that have rapid flicker will display lines with the Auto/Adaptive deinterlacing, the No-interlacing patches are necessary for those.
> 
> Example : Cyclops' ray that is following his eyes in MVC2 or the red lifebar in DOA2. The only thing is that the No-interlacing patch has to be correctly done. It's usually easily implemented in most games but some (Bloody Roar 3) will have many interlacing offset mechanisms.

Of course, that's why I say that, in general, they're no longer necessary and they worsen the experience. The example you're giving would be the 20 or 50 titles, if that many, that really need a patch right now. Compared to the thousands of PS2 games, we're talking about 1% of the catalog.

Immersion95:
> > > Yes, I'm referring to those specific ones, and generally speaking, because PCSX2 already does a good job. You can compare those interlacing patches with them enabled and then disabled with PCSX2's automatic interlacing; you'll see that either there's no difference, or it looks better without those patches. In either case, there's no improvement.
> > 
> > 
> > That's not true for all games. Many games that have rapid flicker will display lines with the Auto/Adaptive deinterlacing, the No-interlacing patches are necessary for those.
> > Example : Cyclops' ray that is following his eyes in MVC2 or the red lifebar in DOA2. The only thing is that the No-interlacing patch has to be correctly done. It's usually easily implemented in most games but some (Bloody Roar 3) will have many interlacing offset mechanisms.
> 
> Of course, that's why I say that, in general, they're no longer necessary and they worsen the experience. The example you're giving would be the 20 or 50 titles, if that many, that really need a patch right now. Compared to the thousands of PS2 games, we're talking about 1% of the catalog.

You could just disable it, there is no worsening.

For fighting games, auto-deinterlacing is generally too distracting. You also need to increase the resolution to at least 4x–6x for the effect to become unnoticeable.
With no-interlacing patches, you only need 2x resolution, and personally, I also prefer a 240p image over 480i in SW.

CRASHARKI:
So, should these patches be added or removed?

Immersion95:
> So, should these patches be added or removed?

You should add it :).

fox140cv:
> So, should these patches be added or removed?

The answer is indicated by the comparison with or without the patch; when there is an improvement, yes, but if there is no improvement, it makes no sense.
In this case, no matter how much I look, I can't see it

with
without

I'm not talking nonsense. It's a simple comparison. One has the interlacing patch enabled in the pull, and the other has it set to automatic, which is the default setting for PCSX2. They're set to 4x to minimize the appearance of the defect, and the screen doesn't vibrate either. These are the facts.

@CRASHARKI do the tests

F0bes:
Any update?

PR #716, opened 2026-05-11
Add missing patches and updates

### LINKED ISSUE #559: GTA:SA, SLES-52541, Patch breaks recording via Pcsx2.
Patch:
"480p Mode"
"Author: PeterDelta"
"SDTV 480p mode at start"

How to reproduce:

Have this patch active at launch and then starting recording using pcsx2s internal recording tool, you get the following message:

"avcodec_open2() for video failed: Invalid argument (-22)"

Image

fox140cv:
Add and update FPS patches, widescreen support, and quality-of-life improvements such as blur removal, etc. Tested without errors

Fixes #559


patches/SLES-50841_93A090CB.pnach
```diff
@@ -0,0 +1,18 @@
+gametitle=Largo Winch - Empire Under Threat (PAL-F) SLES-50841 93A090CB
```
pgert: "PAL-F" -> "PAL-M5".
According to GameIndex and redump:
http://redump.org/disc/12761/

CRASHARKI:
Out of curiosity, why are you changing these codes from patch=1 to patch=0?

fox140cv:
> Out of curiosity, why are you changing these codes from patch=1 to patch=0?

That's how they're published. I'm just updating them.

PR #721, opened 2026-05-19
Call of Duty Finest Hour add missing versions + tidy up

CRASHARKI:
Added and tested the PAL-M and PAL-G versions + tidy up the other versions' patch files.


Mrlinkwii (approved):
looks good , assumed tested

PR #724, opened 2026-05-23
Add and improve patches to Max Payne 2

CRASHARKI:
- Added the widescreen patch made by Arapapa that preserves the original FOV, but added a fix to prevent zoom out during the Graphic Novel parts.
- Added the patches to the missing versions.
- All versions were tested. The Spanish version was played from beginning to end.

### DIFF
--- a/patches/SLES-51954_564F352C.pnach
+++ b/patches/SLES-51954_564F352C.pnach
@@ -3,7 +3,7 @@ gametitle=Max Payne 2 - The Fall of Max Payne (PAL-E) SLES-51954 564F352C
 [Widescreen 16:9]
 gsaspectratio=16:9
 author=PeterDelta, Arapapa & CRASHARKI
-description=Enable enhanced native widescreen.
+description=Enable enhanced native widescreen. (Original widescreen uses vertical cropping; Enhanced expands the horizontal view.)
 patch=1,EE,005996C8,extended,00000010 //00000004
 patch=1,EE,005996D0,extended,00000009 //00000003
 //Zoom fix (Internal widescreen) by Arapapa & CRASHARKI
--- a/patches/SLES-51954_564F352C.pnach
+++ b/patches/SLES-51954_564F352C.pnach
@@ -2,14 +2,19 @@ gametitle=Max Payne 2 - The Fall of Max Payne (PAL-E) SLES-51954 564F352C
 
 [Widescreen 16:9]
 gsaspectratio=16:9
-author=PeterDelta
-description=Enable native widescreen
+author=PeterDelta, Arapapa & CRASHARKI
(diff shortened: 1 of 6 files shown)


PeterDelta:
Good work, I hadn't realized the widescreen needed a fix. How about we change the description to: "Widescreen fix"? Like in the others pnach where we've fixed the problem. I think it's more accurate ;)

CRASHARKI:
> Good work, I hadn't realized the widescreen needed a fix. How about we change the description to: "Widescreen fix"? Like in the others pnach where we've fixed the problem. I think it's more accurate ;)

Well, it's not like it's "broken", so you still did a good job finding the codes to enable the native widescreen. I just happened to come across the patches Arapapa made by accident (I hadn't even realized the game was vertical-), and I thought they were a nice addition to preserve the original FOV. Although, I had to fix the Graphic Novels being zoomed out.

As for the description, if I'm honest, I never really liked the generic "Widescreen fix" comment that patch creators used for pretty much every widescreen-related patch over ten years ago. Now that these descriptions are visible to users, I prefer something more presentable, and I think "Enable enhanced native widescreen" is very accurate since it enables the game's native widescreen functionality while also enhancing it by preserving the original 4:3 FOV.

PeterDelta:
> As for the description, if I'm honest, I never really liked the generic "Widescreen fix" comment that patch creators used for pretty much every widescreen-related patch over ten years ago.

"Widescreen fix" isn't a generic comment, but a precise description, since the original widescreen is cropped compared to how it appears in 4:3.

CRASHARKI:
> "Widescreen fix" isn't a generic comment, but a precise description, since the original widescreen is cropped compared to how it appears in 4:3.

To fix something, it needs to be broken. In this case, it wasn't, it simply had room for improvement. That's why I used the word "enhanced", it took the already working native widescreen and improved upon it.

PeterDelta:
> To fix something, it needs to be broken. In this case, it wasn't, it simply had room for improvement. That's why I used the word "enhanced", it took the already working native widescreen and improved upon it.

You're absolutely right about that. But I understand "fix" in this case as "correction." Perhaps we should describe those patches as "Widescreen correction" to be more precise. Use whatever you like, I was just offering my opinion ;)

RedPanda4552:
"xyz fix" in general is non-desciptive, it clues that something is of course fixed, but lends nothing as to what exactly was the problem, or how it was solved. Without either of those details the comment is essentially useless. It would be strongly preferred to have a short explanation... Taking this example:

> original widescreen is cropped compared to how it appears in 4:3.

This added to the end pretty much gives all the detail you need while still being quick and simple. I'd strongly push for a little added detail like this. Otherwise ten years from now half of this stuff will make no sense to anyone.

PR #726, opened 2026-05-25
Add and review patches for the Hitman Trilogy

CRASHARKI:
Hitman 2: Silent Assassin Changes:
- Added missing patches to the NTSC-U versions.
- General cleanup and tidying.

Hitman: Contracts Changes:
- Added missing patches to multiple versions.
- Removed an incorrect PAL-S CRC file.
- Fixed the 50/60 FPS patch to prevent FMV skipping.
- General cleanup and tidying.

Hitman: Blood Money Changes:
- Added missing patches to multiple versions.
- Improved the Widescreen patch description to clarify its incompatibility with DTV480p.
- Removed the 480p mode patch, as the game already boots into an unskippable menu where players must choose between 50 Hz (PAL only), 60 Hz, and DTV480p. Additionally, encouraging 480p usage via a patch is not a good idea, since that mode breaks Widescreen support.
- General cleanup and tidying.

All tested.

### DIFF
--- a/patches/SLES-50992_5B9ACF79.pnach
+++ b/patches/SLES-50992_5B9ACF79.pnach
@@ -24,4 +24,4 @@ patch=1,EE,20508090,extended,00000001 //00000002
 [NTSC Mode]
 author=Gabominated
 description=NTSC mode at start. FMVs remain in PAL.
-patch=1,EE,003066B4,word,14A20057 //10A20057
+patch=0,EE,003066B4,word,14A20057 //10A20057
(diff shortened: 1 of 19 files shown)


Gabominated:
Please could you update Silent Assassin for me? Change patch=1,EE,003066B4,word,14A20057 //10A20057 to patch=0,EE,003066B4,word,14A20057 //10A20057. (just change patch=1 to patch=0) There's no need to keep rewriting it.

PR #727, opened 2026-05-27
Add patches to Scarface

CRASHARKI:
Changes:
- Add (enable native) widescreen patches.
- Add all patches to "Collector's Edition" version.

All tested.

### DIFF
--- a/patches/SLES-54182_301A1B6E.pnach
+++ b/patches/SLES-54182_301A1B6E.pnach
@@ -1,5 +1,11 @@
 gametitle=Scarface - The World is Yours PAL-M SLES-54182 301A1B6E
 
+[Widescreen 16:9]
+gsaspectratio=16:9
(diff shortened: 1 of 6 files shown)


Gabominated:
There's something strange about the Collector's Edition PAL code and the NTSC version. Do they share the same FPS code? I don't think so, but if they do share the same FPS code, then the only author would be asasega.

CRASHARKI:
> There's something strange about the Collector's Edition PAL code and the NTSC version. Do they share the same FPS code? I don't think so, but if they do share the same FPS code, then the only author would be asasega.

I tested all of them, so the codes are correct. You're more than welcome to run some tests yourself to make sure everything is working properly. Also, the Russian version uses the same addresses, and I agree, I'll credit asasega as the author whenever the codes are the same as theirs.

PR #729, opened 2026-05-30
Improve God of War patches

CRASHARKI:
Changes:
- Added code to enable the game's native widescreen mode where it was missing.
- General cleanup and tidying.

All tested.

### DIFF
--- a/patches/SCES-53133_FB0E6D72.pnach
+++ b/patches/SCES-53133_FB0E6D72.pnach
@@ -1,9 +1,9 @@
-gametitle=God of War (PAL-M) [SCES_531.33] FB0E6D72
+gametitle=God of War (PAL-M) (SCES-53133) FB0E6D72
 
 [Widescreen 16:9]
 gsaspectratio=16:9
 author=Some Chump and PeterDelta
-description=Widescreen fix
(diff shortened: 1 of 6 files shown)


PeterDelta:
By now it's common knowledge that the PS2 cropped widescreen. If you don't like the word "fix," change it to "fixed" or use "corrects the widescreen." I don't see the need for so much redundancy in the description.

CRASHARKI:
> By now it's common knowledge that the PS2 cropped widescreen. If you don't like the word "fix," change it to "fixed" or use "corrects the widescreen." I don't see the need for so much redundancy in the description.

I'm not going to argue about this again. "Widescreen fix" is not a sufficiently descriptive label on its own. My description explains what the patch actually does and how it improves upon the original implementation, which was also suggested by @RedPanda4552.

You've had several opportunities to explain why your description is better, yet the only justification offered so far has been an appeal to "common knowledge". Many users wouldn't even be aware of how the original widescreen mode works, let alone what "Widescreen fix" is supposed to mean.

I have great respect and appreciation for your work, which only makes me more confused as to why this particular wording has become such a point of contention.

crashGG:
Could you apply this modification to God of War (Japan) ?

PR #733, opened 2026-06-08
Updates patches

### LINKED ISSUE #269: GT4 480p in race
I've done some experimenting with Silent's deinterlacing patch which was removed here: https://github.com/PCSX2/pcsx2_patches/commit/18732331eb507c9dc50e45a31c19a2db48efe249

The last line which says "speed timer" slows down the race clock, but doesn't slow down the actual game/race. So effectively you'll be driving 20% faster and lap times will be 17% lower. This is immediately obvious when doing a license test (immediate gold by huge margin). Leaving out that line "fixes" the clock problem. But the game will still be running too quick. I got it to run "properly" (though unfortunately at 50fps) by simply setting the FPS limit at 83.3%. Combined with another deinterlacing patch by Silent, I've now got this:

```pnach
// No interlacing
patch=1,EE,204A2A2C,extended,0000102D

patch=1,EE,2035812C,extended,2442003C // (60 fps?)
patch=1,EE,20A57E70,extended,00000001
```

I had the idea to change the number `3C` to `32`, but that just causes the exact same problem with the clock. So I'm guessing the PAL build simply doesn't support 60FPS (due to tick rate, perhaps?). Unless nobody has found it up until now...

### LINKED ISSUE #657: [BUG] Klonoa 2: background elements in stages/cutscenes load up slowly and "pop in" when widescreen patches are in use
### Describe the Bug

In Klonoa 2: Lunatea's Veil (SLUS-20151), when widescreen patches are enabled, some background elements in stages and cutscenes load with a noticeable delay and visibly "pop in". A particularly clear example can be observed in the Empty Sea of Tears stage, both during the intro cutscene and in gameplay.

### Reproduction Steps

1. Enable "Apply Widescreen Patches" in Settings → Graphics → Display
2. Launch the game
3. Play e.g. Empty Sea of Tears

### Expected Behavior

Background elements should not "pop in" during gameplay or cutscenes.

### PCSX2 Revision

v2.7.43

### System details

- Operating system: Linux (Fedora 43)
- CPU: AMD Ryzen 7 3700X
- GPU: AMD Radeon RX 6800 XT

### Settings

Default settings, but with the "Apply Widescreen Patches" setting enabled.

### Screencasts

**PCSX2 with widescreen patches** (captured on PCSX2 v1.7.3069, but looks the same as in v2.7.43):

https://github.com/user-attachments/assets/60df8b1d-cd0c-495a-b08f-ced7f1b7b1a6

**Real hardware:**

https://github.com/user-attachments/assets/00828001-b109-4033-ac21-14f7e49b0c27

### LINKED ISSUE #659: [BUG] Klonoa 2: wavy text effect renders incorrectly when widescreen patches are in use
### Describe the Bug

In Klonoa 2: Lunatea's Veil (SLUS-20151), when widescreen patches are enabled, the wavy text effect seen on the "Now Loading", "Vision Start", and "Vision Clear" screens renders incorrectly, with parts of the text edges being clipped. This issue occurs on both hardware and software renderers.

### Reproduction Steps

1. Enable "Apply Widescreen Patches" in Settings → Graphics → Display
2. Launch the game
3. Start any stage

### Expected Behavior

The text effect should render correctly and without cutting off.

### PCSX2 Revision

v2.7.43

### System details

- Operating system: Linux (Fedora 43)
- CPU: AMD Ryzen 7 3700X
- GPU: AMD Radeon RX 6800 XT

### Settings

Default settings, but with the "Apply Widescreen Patches" setting enabled.

### Screencasts

**PCSX2 with widescreen patches** (captured on PCSX2 v1.7.3069, but looks the same as in v2.7.43):

https://github.com/user-attachments/assets/ded9049e-dd42-498f-a535-bdb45ca6db0b

**Real hardware:**

https://github.com/user-attachments/assets/248aaed6-2558-4851-933f-58b2aa280972

PeterDelta:
I've updated several patches with minor fixes and more precise descriptions. Most of the games have been completed or are well into their story. The most relevant changes are:

- Klonoa 2: Fixed rendering of environments and text clipping.

- Dead or Alive 2: Added widescreen in PAL mode.

- Midnight Club II: Minor fixes for animations.

- Gran Turismo 4: WS: It is not necessary to force 16:9 at the start because the game saves this configuration to the memory card and loads it before starting the game. Therefore, there's no need to add extra workload that the game already handles. I've also added 480p and 1080i patches for the PAL version with all fixes. I've completed the game and am around 70% finished, but I've tested all the tracks. There shouldn't be any errors.

- Tourist Trophy: Added 480p and 1080i patches for the PAL version. I completed it 100% without errors. For the same reason as in GT4, I don't force widescreen at the start, the settings are saved to the memory card.

- Rally Fusion: The widescreen patch had numerous problems, all have been fixed.

- Half-Life: The old widescreen patch modified the 4:3 aspect ratio. This change leaves it intact and only modifies the 16:9 aspect ratio, which is the one needed. I also completed it from beginning to end without errors.

- Crazy Taxi: The Japanese widescreen version had errors, these have been fixed.

- Spider-Man: Widescreen with all levels corrected

Add:
- NBA Series: Some games had widescreen patches, but they suffered from rendering errors. I've fixed them all and added the entire NBA series.

- Add WS: Tiger Woods PAL

Fixes #269
Fixes #657
Fixes #659


CRASHARKI:
> Gran Turismo 4: WS: It is not necessary to force 16:9 at the start because the game saves this configuration to the memory card and loads it before starting the game. Therefore, there's no need to add extra workload that the game already handles. I've also added 480p and 1080i patches for the PAL version with all fixes. I've completed the game and am around 70% finished, but I've tested all the tracks. There shouldn't be any errors.

Many games save the widescreen setting on the memory card, but that's not a reason to remove the patch. This patch is no different from many others that exist specifically to automate widescreen functionality.

I can't help but feel that this may have been done with malicious intent given our recent disagreements, but I'd like to be proven wrong, so please restore the Widescreen patch exactly as it was. This change merely undoes the work we've put into improving the user experience.

The same applies to the other games whose widescreen patches were fixed but widescreen mode was left disabled. Having a widescreen patch that does not actually enable widescreen functionality is unnecessarily inefficient and defeats the purpose of the patch.

Lastly, "Corrects display cropping to 16:9" does not make sense, even on a grammatical level. Furthermore, "Enable enhanced native widescreen. (Original widescreen uses vertical cropping; Enhanced expands the horizontal view.)" is simply a more accurate description, at least unless one decides that a patch literally called "Widescreen 16:9" should not, for some reason, enable the game's native widescreen mode.

CRASHARKI:
I just tested the new Simpsons Hit & Run patch. The Multiplayer section definitely seems improved, but now the Main Menu and the shop screens appear narrower. Why not combine the best parts of both versions into a more complete patch?

patches/SLES-50997_55E708DA.pnach
```diff
@@ -1,25 +1,17 @@
...
+author=PeterDelta & Arapapa
+description=Renders the game in 16:9 aspect ratio.
+patch=0,EE,001AE5A0,word,3C0140AB
+patch=0,EE,001066E0,word,3C013F06
+patch=0,EE,001066E4,word,34216666
+patch=0,EE,002F9CE8,word,3C013FAB
+
+[NTSC Mode]
+author=Gabominated
+description=Modo NTSC al iniciar.
```
pgert: description=NTSC mode at start.

patches/SCUS-97328_77E61C8A.pnach
```diff
@@ -2,23 +2,8 @@ gametitle=Gran Turismo 4 [SCUS-97328] (U)
 
 [Widescreen 16:9]
 gsaspectratio=16:9
-author=CRASHARKI & Aero_
-description=Run the game at Enhanced 16:9 Widescreen Aspect Ratio from the start.
-
-// Enable Widescreen from the start by CRASHARKI
```
pgert, line 8: If you insist on having this removed, why not make it a hack of its own:
"[Autoboot into 16:9 mode]".

patches/SCES-51719_44A61C8F.pnach
```diff
@@ -1,23 +1,9 @@
-gametitle=Gran Turismo 4 [SCES-51719] (PAL-M)
+gametitle=Gran Turismo 4 (PAL-M) SCES-51719 44A61C8F
 
 [Widescreen 16:9]
 gsaspectratio=16:9
-author=CRASHARKI & Aero_
-description=Run the game at Enhanced 16:9 Widescreen Aspect Ratio from the start.
-
-// Enable Widescreen from the start by CRASHARKI
```
pgert, line 8: If you insist on having this removed, why not make it a hack of its own:
"[Autoboot into 16:9 mode]".

patches/SLES-53557_692CBA8E.pnach
```diff
@@ -13,13 +13,19 @@ description=Disables the brownish yellow post processing filter
 patch=1,EE,00551058,extended,20000001
 
 [NTSC Mode]
-author=PeterDelta
+author=Felixthecat1970
```
pgert, line 16: "Felixthecat1970" -> "felixthecat1970".

patches/SCUS-97328_77E61C8A.pnach
CRASHARKI: The idea is to have a global enabler, so doing it that way would be inefficient. The patch is called "Widescreen 16:9", so that's exactly what it should do: make the game run in a 16:9 widescreen aspect ratio.

patches/SCES-51719_44A61C8F.pnach
CRASHARKI: The idea is to have a global enabler, so doing it that way would be inefficient. The patch is called "Widescreen 16:9", so that's exactly what it should do: make the game run in a 16:9 widescreen aspect ratio.

PeterDelta: https://github.com/user-attachments/assets/55abcb60-3bf2-412c-9782-5b82ccc839a1

I think it would be better to use separate files if you want to add it. The game loads the 16:9 configuration before starting, as you can see in the video, so the code just creates unnecessary workload. Furthermore, the patch is poorly designed.

PeterDelta:
> I just tested the new Simpsons Hit & Run patch. The Multiplayer section definitely seems improved, but now the Main Menu and the shop screens appear narrower. Why not combine the best parts of both versions into a more complete patch?

Our differing opinions are forgotten; they're just opinions ;). 

This is how it has to look for the whole game to display perfectly in 16:9. To get to this point, I've had to make numerous corrections to avoid errors in the HUD, as it's very problematic, changing one thing affects the other.

CRASHARKI:
> This is how it has to look for the whole game to display perfectly in 16:9. To get to this point, I've had to make numerous corrections to avoid errors in the HUD, as it's very problematic, changing one thing affects the other.

Oh, yeah, no doubt the HUD is super unstable, I noticed that too when messing with the values.

I just tested your patch with the addition of this line:
`patch=1,EE,0031eea4,word,00000000 // 10400004 Main Menu`

It fixed the issues I mentioned in the Main Menu and shop screens, and I didn't notice any new problems at all. Is there any particular situation where this code might cause issues? If so, I can test it.

Otherwise, this single line combined with your latest changes makes the patch an objective improvement over the version currently in the repository, which is great.

CRASHARKI:
> I think it would be better to use separate files if you want to add it.

I don't see how that would work, or fully understand what you mean. The patch needs to be in the right place to be enabled by the global option.

> The game loads the 16:9 configuration before starting

The game doesn't automatically start in widescreen as, unfortunately, it doesn't read the BIOS configuration. If you'd like to verify this, remove your save data and launch the game with all patches disabled; you'll get a stretched 4:3 image. Given that behavior, the patch is justified in the same way as widescreen patches for other games in this situation.

> Furthermore, the patch is poorly designed.

This isn't my prettiest patch. However, despite the implementation being somewhat rough, it is functional. I think it would be more productive to improve the implementation while preserving the functionality rather than removing it entirely.

Regarding the performance concerns, the impact appears to be negligible in practice. These patches have worked this way for a long time without any noticeable issues.

More broadly, I'm concerned about the apparent shift away from enabling widescreen mode in widescreen patches. Since the purpose of these patches is to provide a proper widescreen experience, I think preserving that functionality should remain a priority.

fox140cv (approved):

PR #739, opened 2026-06-28
SLES-53096: Worms 4 Mayhem 50 FPS patch

wormi-ra:


Gabominated:
Hey bro, actually it can be done without adjusting the speed, like Worms Forts - Under Siege [https://github.com/Gabominated/PCSX2/blob/main/PCSX2%20Patches/SLES-53096_DC85FC8F.pnach](https://github.com/Gabominated/PCSX2/blob/main/PCSX2%20Patches/SLES-53096_DC85FC8F.pnach)

wormi-ra:
Hey, very nice, I assumed the address I found was the the same equivalent to the one you've used in Worms Forts, I didn't thought there could be multiple addresses influencing the frame rate.
However I have an issue when running the game with your patch, the game is stuck to a black screen after the intro cutscenes. I think the patch only works when applied after the game is started.
It also breaks after watching any story cutscene, also the loading splashscreen doesn't appear during transitions

Gabominated:
> Hey, very nice, I assumed the address I found was the the same equivalent to the one you've used in Worms Forts, I didn't thought there could be multiple addresses influencing the frame rate. However I have an issue when running the game with your patch, the game is stuck to a black screen after the intro cutscenes. I think the patch only works when applied after the game is started. It also breaks after watching any story cutscene, also the loading splashscreen doesn't appear during transitions

ok, try again, Apparently, the loading screen needs to be conditioned to avoid other issues.

wormi-ra:
I figured out the assembly instruction that was reading this address in the main loop and updated my patch to change this assembly instruction to always read `0x0001` instead, works perfectly between loading screens.
Now only one patch at boot time is applied

Gabominated:
> I figured out the assembly instruction that was reading this address in the main loop and updated my patch to change this assembly instruction to always read `0x0001` instead, works perfectly between loading screens. Now only one patch at boot time is applied

I think that new code isn't right; everything is still at 25fps.

wormi-ra:
Nevermind, I messed up my patches

wormi-ra:
Your updated patch works fine, we can use yours instead

Gabominated:
> Your updated patch works fine, we can use yours instead

Alright bro, good to know.

PR #740, opened 2026-06-28
Update SCUS-97481_2F123FD8.pnach

wxvu:
Adds a patch to force-enable 'High-Res Mode' for _God of War II_ at startup, bypassing the need to input a hidden code.

image
image


Mrlinkwii:
@wxvu  can this be rebased ?

PR #744, opened 2026-07-05
Update SCUS-97128_4167D813.pnach

mrnecromancer247:
Adds a fix for white pixel lines that appear around the health, arrows and mana HUD elements when the internal rendering resolution is increased above native.

Tested:
- Vulkan
- D3D12
- Native, 2x, 4x and 6x internal resolution

Result:
- White HUD artifacts are removed.
- No visible issues observed during gameplay

Before
image

After
image


pgert:
Did you port my hack?
Then give me credit, will you.

author=pgert, ported by mrnecromancer247

mrnecromancer247:
Oh, hello there, pgert!
I definitely used your patch as a starting point. Yeah, the values match to the byte — I was searching for the right addresses on the NTSC version and played around with the debugger a bit. Guess you can call it a "port."
I'll update the author line to give you credit.

PR #745, opened 2026-07-12
ci: update workflows to hopefully get rid of intermittent failures

xTVaser:
Does a few things:
- Moves the release step out of the workflow that runs on PRs and instead into a workflow that only runs on pushes to `main`
- Adds some retries on the release creation step incase github's API is still eventually consistent.  Should hopefully resolve errors like this https://github.com/PCSX2/pcsx2_patches/actions/runs/28344568683/job/84135110444
  - Broke up the commands with logs in between so it's easier to tell which command specifically failed in the future.
- However another reason why the above could have failed in the past is because there's a race condition where simultaneous builds are all fighting over the same tag, that is solved via the new `concurrency` addition where only a single build can be active at a time, and it will always be the most recent.

Should work but of course it's a CI change so, kinda have to merge to see

### DIFF
--- a/.github/workflows/pack_patches.yml
+++ b/.github/workflows/pack_patches.yml
@@ -1,6 +1,6 @@
 name: Pack patches
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
   pack:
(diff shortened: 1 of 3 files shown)


F0bes (approved):
Nice, didn't know of `cancel-in-progress`

PR #754, opened 2026-07-29
Need for Speed Carbon: Add exclusive cars unlock patch

NicknineTheEagle:
This patch for Need for Speed: Carbon unlocks 12 cars that are not normally available in PS2 version.

These cars were only released for PC and X360 as DLC.
* Porsche 911 Turbo
* Dodge Challenger R/T
* Toyota Corolla GT-S
* Infiniti G35
* Shelby GT500 '06
* Pagani Zonda F

These cars are online rewards. Since old-gen versions don't have online multiplayer, they're not normally available outside of Challenge Series.
* Porsche 911 GT3 RS
* Lotus Europa S
* Pontiac GTO
* Lexus IS300
* Toyota MR2
* Plymouth Roadrunner

Only 911 Turbo, Challenger and G35 are usable in Career mode, the rest are only usable in Quick Race mode.

Originally made for SLES-54321, but it's also reported to work on SLES-54492, SLES-54493, SLUS-21493, SLUS-21494.


PeterDelta:
This seems more like cheats than patches to me. But that's just my humble opinion.

F0bes:
meh, i'll let it slide, don't use this as a precedent though

F0bes:
Sorry, I ended up reverting this before it gets bundled with the emulator.

@NicknineTheEagle I'm concerned about this being allowed in RA hardcore mode. Does this patch give you an unfair advantage?

NicknineTheEagle:
> Sorry, I ended up reverting this before it gets bundled with the emulator.
> 
> @NicknineTheEagle I'm concerned about this being allowed in RA hardcore mode. Does this patch give you an unfair advantage?

Hmm, not sure? Only three cars from this patch can even be used in Career mode. You do get them for free because of how DLC unlocks work, but none of them are OP, the best car in the game is still Corvette Z06 by far.

F0bes:
I'll ask the RA people...

F0bes:
Yeah, according to their hardcore requirements we should only allow QoL patches.

We were talking a bit about having a `cheats` folder (this repo only has a `patches` folder) that could allow stuff like this. We would have to figure out how the review process would go.

The issue is, everything in this repo is allowed to be used in hardcore mode as of now.

NicknineTheEagle:
Bummer. I'll hold on to the patch for the time being, if PCSX2 gets a separate cheats repo in the future, I'll submit it there.

PR #758, opened 2026-08-06
Rogue Galaxy (SCES-54552): Remove Blackbars patch now covers all cutscenes

caribbeansc:
The current **Remove Blackbars** patch zeroes the 1.0 constant at a single call site of the game's letterbox setter function (`0x279DB0`), but the game raises the bars from four different call sites (`0x244854`, `0x261114`, `0x261FB4`, `0x27B25C`), so black bars still appeared in many in-engine cutscenes.

This PR patches the common code instead of one call site:

- `0x279DE0`: `mov.s f22,f12` → `mtc1 zero,f22` — forces the letterbox target amount to 0 for every caller of the setter.
- `0x279F20`: `bc1t` → `b` — the letterbox draw routine unconditionally takes its existing early-out (the branch it already takes when the bar amount is 0), so the bars are never rendered even if the letterbox state was already raised.

Calls that lower the bars and screen fades use separate paths and are unaffected. Tested on PAL (SCES-54552), including cutscenes where the previous patch had no effect.

Credit kept for El_Patas (original patch author).

https://claude.ai/code/session_01271m1NhungjpvuciVycw5M

I used an LLM to help me find the right functions. I tested all by myself manually.


F0bes:
Another person is gonna have to test this. I do not trust an AI to figure out what to patch.

F0bes:
Also, please turn off the [automatic copilot review](https://docs.github.com/en/copilot/how-tos/copilot-on-github/set-up-copilot/configure-automatic-review) stuff if you're going to contribute here.

What a waste of energy it was for it to literally tell me what you already said.

patches/SCES-54552_CBB4B383.pnach
```diff
@@ -15,9 +15,10 @@ patch=1,EE,001D4E2C,word,24060000 //02C23021 Top
 patch=1,EE,001D4E7C,word,240201C0 //00451021 Bottom
 
 [Remove Blackbars]
-author=El_Patas
+author=El_Patas, caribbeanwebdev
```
pgert, line 18: author=El_Patas, Claude AI & caribbeanwebdev

PeterDelta:
The problem I see with this patch is that it resets the function to 0. It might work, but since it was done by an AI, you have to be 100% sure that those functions don't control another game variable, because if they do, it's ignored or can cause glitches.

If you say there are scenes where the black bars are missing, the best thing to do is find the addresses that control those scenes, just like the current address does, which is only responsible for that.

patches/SCES-54552_CBB4B383.pnach
caribbeansc: Legit, send the changes request 😂

caribbeansc:
> The problem I see with this patch is that it resets the function to 0. It might work, but since it was done by an AI, you have to be 100% sure that those functions don't control another game variable, because if they do, it's ignored or can cause glitches.
> 
> If you say there are scenes where the black bars are missing, the best thing to do is find the addresses that control those scenes, just like the current address does, which is only responsible for that.

Thanks for the feedback. I'm doing a full replay of the game right now, so I'll be able to notice if there are any anomalies, but I'll try the approach you mentioned and let you know.

caribbeansc:
> Another person is gonna have to test this. I do not trust an AI to figure out what to patch.

I've already tested it quite a bit, but it would be very helpful if someone else could try it. I just updated the PR description to clarify that I used an LLM to help me find the correct function.

I hope this patch update is useful to someone and helps them enjoy this amazing game even more, if not, no problem.
I didn't even know it had GitHub Copilot for reviews. I've already disabled it. Thanks for letting me know.

PR #761, opened 2026-08-30
Add Fatal Frame viewfinder stick swap patches

DAOWAce:
First time contributor, sorry if I'm doing anything wrong (let me know).

This adds a viewfinder stick swap patch for the NTSC-U, PAL, and NTSC-J releases. (made for NA, ported to others)

[Long requested feature](https://forums.pcsx2.net/Thread-Fatal-Frame-Tecmo-s-troublesome-analog-stick-layout) that only ever got [addressed for the JP release](https://forums.pcsx2.net/Thread-Update-1-03-PS4-controller-and-Fatal-Frame-1-Project-Zero-analog-stick-solution) in an older version of PCSX2.

I didn't play through the entire game yet, but tested it working (including trying to break it) on every control scheme. Also had one person test the PAL version and [confirm it working](https://www.reddit.com/r/fatalframe/comments/xneful/comment/p2jxbqy/)

### DIFF
--- a/patches/SLES-50821_22E91837.pnach
+++ b/patches/SLES-50821_22E91837.pnach
@@ -140,3 +140,11 @@ patch=1,EE,0035905C,word,00000000
 // It also wracks up the mirror rendering in the Cutscenes:
 //  https://github.com/PCSX2/pcsx2_patches/pull/588
 // ==========
+
+[Viewfinder Stick Swap]
+author=DAOWAce
+description=Swaps the analog stick functions while using the Camera Obscura viewfinder.
+
+patch=1,EE,2018CE4C,extended,90620193
+patch=1,EE,2018CE60,extended,90620194
+patch=1,EE,2020DD98,extended,24040000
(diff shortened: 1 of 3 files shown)


pgert:
Is this hack intended for DS4Windows and other DS4 wrappers?
Because with DS2 & DS3 there is no problem; the hack just messes things up.
If so, (as a suggestion) add "Intended for DS4 wrappers." in description.

MrWrightyLC:
@pgert What are you talking about, Have you played Fatal Frame 1? If so, you'd know that this patch changes the joystick functions when using the camera obscura. 

> Fatal Frame (1) has a truly bizarre analog stick setup.
While navigating the mansion, you'll use the Left Analog Stick to move your character, and the Right Analog Stick to aim your flashlight.
Very standard, most games will use a similar setup. However....
When you pull out (tap Circle by default) your camera in order to fight the ghosts attempting to harm you, the Left Analog Stick becomes the camera aim, and the Right Analog Stick moves your character instead.

> Fatal Frame 2 and 3, allow you to choose which Analog Stick setup you'd like use.
But not Fatal Frame (1); you're absolutely stuck with this awful setup.
Even with muscle memory, the constant stick function swamping depending if you're moving in 3rd or 1st person (camera mode) gets old really fast.

MrWrightyLC (approved):
Tested and Working (Pal version)
I finished the introduction and the first night; I also tested it with the different control schemes, and it works great.
It's a nice QoL improvement!

PR #762, opened 2026-08-31
NeoGeo Battle Coliseum: Add loading time fixes for PAL and NTSC-J

Immersion95:
## Summary

Adds loading time fixes for the PAL (`SLES-54395`) and NTSC-J (`SLPS-25558`) versions of NeoGeo Battle Coliseum.

These versions can experience abnormally long or seemingly endless loading screens. The game can treat a temporarily unavailable CDVD status as a real read error and repeatedly restart the same read.

The patches make the game wait and check again when the status is temporarily unavailable, while keeping the original retry behavior for real CDVD errors.

## Files

* `SLES-54395_FE883E27.pnach`
* `SLPS-25558_7D6924E8.pnach`

Related: PCSX2/pcsx2#11624


pgert:
The "comment" entries might cause problems.
As a suggestion, make it like this:

[Loading Time Fix]
author=Immersion95
description=Prevents unusually long or endless loading screens.
// Simple explanation: when the game cannot check the disc status immediately,
//  it now waits instead of restarting the same read over and over.
// Technical details: sceCdGetError() is handled as three different results.
// -1 waits, 0 completes the load, and real positive CDVD errors
//  keep the original retry behavior.

Immersion95:
Done :)

PR #764, opened 2026-09-02
Add No Old Movie Filter for Brothers in Arms - Earned in Blood (SLUS-21310)

Teddy101991:
A simple Patch that disables the old movie filter. I played through a few missions completely and it works great.

**Before:**
Brothers in Arms - Earned in Blood_SLUS-21310_20260902164007

**After:**
Brothers in Arms - Earned in Blood_SLUS-21310_20260902164013


Mrlinkwii (approved):
assumed was tested

PR #766, opened 2026-09-07
Fix Code Veronica X startup skip system-data loading

JoeKoh27:
## Problem

The current "Skip Startup logos and warning" patch changes the initial startup state from 0x00300002 to 0x00300008.

This skips the startup path containing the system-data/memory-card loading sequence. As a result, game unlocks stored in the system save, such as Battle Game and Battle Game characters, are not loaded after a console reset while the patch is enabled.

## Fix

This replacement preserves the normal startup and system-data loading flow.

After system-data loading has completed, pressing or holding START skips the remaining startup warning/logo sequence and proceeds to the title screen.

The injected code uses unused linker padding at 0x00301F14-0x00301F7F in SLUS-20184.

## Tested

- NTSC-U SLUS-20184, CRC 24036809
- Current patch: Battle Game unlock is missing after console reset
- Patch disabled: Battle Game unlock loads normally
- Replacement patch: Battle Game unlock loads normally
- START skips remaining startup sequence after memory-card processing
- Normal startup remains functional when START is not pressed


JordanTheToaster:
New contributors are not allowed to use AI and especially not on patches. Please see the PCSX2 website for guidelines.

PR #768, opened 2026-09-09
FMV skip patches for Sonic Unleashed

FrankyBuster:
These patches allow for skipping fmvs in Sonic Unleashed anytime, the original PAL version of the patch was made by xan1242 and i ported it to the NTSC ver.

https://github.com/user-attachments/assets/8a8434c3-db83-4149-8b94-57551db89da5

Normally you can't skip cutscenes in this game unless it considers them as already watched on that save or during the same session, speedrunners would use completed saves to watch all the movies in the gallery to then make a new save that's capable of skipping cutscenes during the same session, these patches skip that process so you can just skip them anytime


pgert:
Avoid potential problems by changing "comment" into "description".
"description=Allows you to always skip FMVs by pressing Start."

Issue #767, opened 2026-09-08
[add] widescreen patch The Suffering PAL (SLES-51693)


geextahslex:
```
gametitle=The Suffering PAL (SLES-51693) FB680748
[Widescreen 16:9]
gsaspectratio=16:9
author=the_deep2
description=Renders the game in 16:9 aspect ratio

patch=1,EE,001E766C,word,E7B40010
patch=1,EE,001E7670,word,8E020338
patch=1,EE,001E7674,word,AE020330
patch=1,EE,001E7678,word,8E02033C
patch=1,EE,001E767C,word,AE020334
patch=1,EE,001E7680,word,C6030338
patch=1,EE,001E7684,word,468018E0
patch=1,EE,001E7688,word,3C013F00
patch=1,EE,001E768C,word,4481B000
patch=1,EE,001E7690,word,C600033C
patch=1,EE,001E7694,word,46800020
patch=1,EE,001E7698,word,3C014500
patch=1,EE,001E769C,word,44811000
patch=1,EE,001E76A0,word,3C0142AB
patch=1,EE,001E76A4,word,44810800
patch=1,EE,001E76A8,word,44862000
patch=1,EE,001E76AC,word,46802120
patch=1,EE,001E76B0,word,461618C2
patch=1,EE,001E76B4,word,3C014300
patch=1,EE,001E76B8,word,44813000
patch=1,EE,001E76BC,word,46160002
patch=1,EE,001E76C0,word,E60102E0
patch=1,EE,001E76C4,word,44852800
patch=1,EE,001E76C8,word,46802960
patch=1,EE,001E76CC,word,3C01BF00
patch=1,EE,001E76D0,word,44813800
patch=1,EE,001E76D4,word,460310C1
patch=1,EE,001E76D8,word,AE050328
patch=1,EE,001E76DC,word,46001081
patch=1,EE,001E76E0,word,AE06032C
patch=1,EE,001E76E4,word,C6010330
patch=1,EE,001E76E8,word,46800860
patch=1,EE,001E76EC,word,8E05034C
patch=1,EE,001E76F0,word,C6000334
patch=1,EE,001E76F4,word,46800020
patch=1,EE,001E76F8,word,3C013C8E
patch=1,EE,001E76FC,word,3421FA36
patch=1,EE,001E7700,word,4481B800
patch=1,EE,001E7704,word,460518C0
patch=1,EE,001E7708,word,46041080
patch=1,EE,001E770C,word,46000B06
[...]

PR #771, opened 2026-09-11
The Suffering (SLES-51693_FB680748 & SLES-53626_AC4B758C)

pgert:
Addresses #767 & #773.


PeterDelta:
c

It was just mentioned that there is a problem with the widescreen for AC4B758C. Why are you adding that same patch without verifying it and including irrelevant comments?

pgert:
> It was just mentioned that there is a problem with the widescreen for AC4B758C. Why are you adding that same patch without verifying it and including irrelevant comments?

I got confused, sorry.

